PR #1, opened 2018-09-24
APM-1079 & APM-1484: Add initial Implementation for java exception catching

npalaska:
- Gradle supported files added to the project for easier dependency management 
- Unhandled exception catching by registering a custom uncaught handler
- Messagepack Implementation to wrap the stacktrace information
- Device registration implementation


bleib1dj:
@npalaska by using gradle does that mean customers will be forced to use gradle?

.idea/modules/agent/agent.iml
```diff
@@ -0,0 +1,12 @@
...
+  <component name="NewModuleRootManager" inherit-compiler-output="true">
+    <exclude-output />
+    <content url="file://$MODULE_DIR$/../../../agent">
+      <excludeFolder url="file://$MODULE_DIR$/../../../agent/.gradle" />
+      <excludeFolder url="file://$MODULE_DIR$/../../../agent/build" />
+    </content>
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+</module>
```
bleib1dj: .idea dir should be added to the .gitignore file

gradle.properties
```diff
@@ -0,0 +1,3 @@
+theGroup=io.auklet
+theName=agent-java
+theVersion=local-build
```
bleib1dj: @rjenkinsjr are there best practices for the naming in this properties file?

README.md
```diff
@@ -1 +1,34 @@
...
+2. Initialize the auklet agent at the very beginning of `main` using app_id and api_key as following:
+```
+import io.auklet.agent.Auklet;
+   
+public class demo {
+   
+   public static void main(String []arg) {
+       Auklet.init(<App_Id>, <Api_key>);
+       ...
+       }
```
bleib1dj: @npalaska just formatting but I think this `}` should be deindented by 4 spaces.

.idea/modules/agent/agent.iml
bleib1dj: @npalaska can you please add the `.idea` directory to the gitignore file

README.md
```diff
@@ -1 +1,34 @@
...
+
+# Features 
+
+1. Automatic report of unhandled exceptions.
+2. Location, and system metrics identification for all issues.
+
+
+# Quickstart
+
+1. Download the agent jar file and import it in the java application.
```
bleib1dj: @npalaska where will they download the jar from? Do we have it hosted somewhere? Do they get it from this repo? Can we provide a hyperlink?

.idea/modules/agent/agent.iml
shogun656: What @bleib1dj said. This dir should be in the `.gitignore`

gradlew
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env sh
+
+##############################################################################
+##
+##  Gradle start up script for UN*X
```
shogun656: do you mean `UNIX`?

settings.gradle
```diff
@@ -0,0 +1,4 @@
+rootProject.name = theName
+include 'agent'
+include 'agent'
```
shogun656: do you need `include 'agent'` twice here?

src/main/java/io/auklet/agent/Auklet.java
```diff
@@ -0,0 +1,21 @@
...
+import org.eclipse.paho.client.mqttv3.MqttClient;
+
+public class Auklet {
+
+    static protected String AppId;
+    static protected String ApiKey;
+    static protected String baseUrl = "https://api-staging.auklet.io/";
+    static protected MqttClient client;
+
+    public static void init(String appId, String apiKey){
```
shogun656: You should also add the baseUrl as a parameter to the init function set it to default to the produnction url. Since Java doesn't do default parameter values (come on java), I suggest either using one of these ways to create a default parameter https://stackoverflow.com/a/19719701/6709710

README.md
rjenkinsjr: @bleib1dj @npalaska we'll need to publish it to JCenter or Maven Central so that people can easily include it in their app.

bleib1dj: @rjenkinsjr makes sense. Is that a manual process or something like pypi that we should handle in the build via another PR?

README.md
```diff
@@ -1 +1,34 @@
-# Auklet-Agent-Java
\ No newline at end of file
+# Auklet for Java
+
+This is the official Java agent for Auklet. It officially supports Java 9+.
```
rjenkinsjr: Are we supporting Java 8?

src/main/java/io/auklet/agent/AukletUnhandledException.java
```diff
@@ -0,0 +1,93 @@
...
+        this.defaultExceptionHandler = defaultExceptionHandler;
+    }
+
+    @Override
+    public void uncaughtException(Thread thread, Throwable thrown) {
+
+        if (defaultExceptionHandler != null) {
+            // call the original handler
+            defaultExceptionHandler.uncaughtException(thread, thrown);
+            System.out.println("We are here");
```
shogun656: We should get rid of all the prints when the PR is ready

README.md
```diff
@@ -1 +1,34 @@
...
+It is currently tested on Oracle java Hotspot JVM. We plan on doing more 
+testing on other currently active JVMs.
+
+
+# Features 
+
+1. Automatic report of unhandled exceptions.
```
rjenkinsjr: For consistency let's make these non-numbered bullets.

src/main/java/io/auklet/agent/AukletUnhandledException.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        else if (!(thrown instanceof ThreadDeath)) {
+            List<Object> list = new ArrayList<>();
+            //MessageBufferPacker packer = MessagePack.newDefaultBufferPacker();
```
shogun656: We should get rid of all the uncommented code as well when the PR is ready

src/main/java/io/auklet/agent/AukletUnhandledException.java
```diff
@@ -0,0 +1,93 @@
...
+            // CHECKSTYLE.OFF: RegexpSinglelineJava
+            System.err.print("Exception in thread \"" + thread.getName() + "\" ");
```
shogun656: I'm guessing `System.err.print` is the way to log in java?

build.gradle
```diff
@@ -0,0 +1,47 @@
...
+apply plugin: 'whitesource'
+
+group "${theGroup}"
+version "${theVersion}"
+
+repositories {
+    mavenCentral()
+}
+dependencies {
+    compile "org.apache.kafka:kafka-clients:0.11.0.0"
```
rjenkinsjr: Is this still needed?

README.md
npalaska: @bleib1dj We dont have it hosted atm. However, I can check the jar file in this repo if you want and so people can just download the jar from here but not sure that's the best way to do it, maybe @rjenkinsjr could provide some input. Maybe we can store it in S3 and give a hyperlink from where they can download it.

src/main/java/io/auklet/agent/AukletUnhandledException.java
```diff
@@ -0,0 +1,93 @@
...
+            thrown.printStackTrace(System.err);
+
+            System.out.println("library Uncaught Exception caught from app  " + thrown.getMessage());
+            System.out.println("Uncaught Exception stacktrace is ");
```
shogun656: these should be logs and not print

README.md
bleib1dj: @npalaska yeah @rjenkinsjr and I are discussing it offline. This will be outside the scope of this PR.

gradle.properties
rjenkinsjr: Group is `io.auklet` as it should be. Version is `local-build` so that our auto-versioning can override it, which is also correct.

Not sure yet about `theName`, which translates to the Maven artifact ID when we eventually start publishing. The full artifact coordinate would be `io.auklet:agent-java:VERSION`. The artifact ID is really a function of marketing IMO.

Saying `agent-java` is A. a bit redundant since it will be a JAR file, which would tip anyone off right away, and B. might give the impression that we're exclusionary to other JVM languages like Scala. Not sure if that's the direction we want to go. For now it's fine, but we'll need to decide before we start doing publishing.

settings.gradle
rjenkinsjr: Why is this line included twice?

src/main/java/io/auklet/agent/Auklet.java
```diff
@@ -0,0 +1,21 @@
+package io.auklet.agent;
+
+import org.eclipse.paho.client.mqttv3.MqttClient;
+
+public class Auklet {
```
rjenkinsjr: If you don't intend for this class to be extended, it should be `final` and its member variables should be `private`.

src/main/java/io/auklet/agent/AukletUnhandledException.java
```diff
@@ -0,0 +1,93 @@
...
+
+import org.eclipse.paho.client.mqttv3.MqttException;
+import org.eclipse.paho.client.mqttv3.MqttMessage;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class AukletUnhandledException implements Thread.UncaughtExceptionHandler {
```
rjenkinsjr: This class should be named `AukletExceptionHandler` or something similar. When a Java class name ends with `Exception` it's assumed that it is an exception type, which is not the case here.

Also, if you don't intend for this class to be extended, it should be `final` and its member methods should be either `private` or `public`.

src/main/java/io/auklet/agent/Device.java
```diff
@@ -0,0 +1,184 @@
...
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
+import java.io.*;
+
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.util.Scanner;
+
+public class Device {
```
rjenkinsjr: If you don't intend for this class to be extended, it should be `final`.

Also, every member in this class is `static`, which implies that the class should be `static` too...but is that really what you want, or should this class be instantiable so that it represents a device?

src/main/java/io/auklet/agent/MQTT.java
```diff
@@ -0,0 +1,134 @@
...
+import javax.net.ssl.TrustManagerFactory;
+import java.io.BufferedInputStream;
+import java.io.FileInputStream;
+import java.nio.charset.StandardCharsets;
+import java.security.KeyStore;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.util.Scanner;
+
+public class MQTT {
```
rjenkinsjr: If you don't intend for this class to be extended, it should be `final`.

src/main/java/io/auklet/agent/Device.java
```diff
@@ -0,0 +1,184 @@
...
+            file.write(obj.toJSONString());
+            file.flush();
+
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+
+    }
+
+    public static String getClient_username(){
```
shogun656: We should be using Camel Case for java. Not a mixture of both

src/main/java/io/auklet/agent/Messages.java
```diff
@@ -0,0 +1,37 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.json.simple.JSONObject;
+import org.msgpack.jackson.dataformat.MessagePackFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+public class Messages {
```
rjenkinsjr: If you don't intend for this class to be extended, it should be `final`.

src/main/java/io/auklet/agent/util.java
```diff
@@ -0,0 +1,99 @@
...
+import java.net.InetAddress;
+import java.net.NetworkInterface;
+import java.net.SocketException;
+import java.net.UnknownHostException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class util {
```
rjenkinsjr: If you don't intend for this class to be extended, it should be `final` and its member variables should be either `private` or `public`.

Also, it seems like there are other classes that only provide utility functionality because they are `static`, like `MQTT` and `Messages`. We need to clearly delineate which aspects of the agent should be represented as objects and which should be represented as utility classes.

build.gradle
npalaska: @rjenkinsjr no I am cleaning it up right now

gradle.properties
bleib1dj: @rjenkinsjr in regards to `agent-java` what would you recommend?

settings.gradle
npalaska: @rjenkinsjr not sure, didnt realize it

src/main/java/io/auklet/agent/Auklet.java
npalaska: @rjenkinsjr gotcha

src/main/java/io/auklet/agent/AukletUnhandledException.java
npalaska: Yeah you are right the naming conventions is wrong

src/main/java/io/auklet/agent/util.java
npalaska: @rjenkinsjr yes I am changing all the public classes to `protected` except the Auklet class so that they wont be visible outside our package

src/main/java/io/auklet/agent/Auklet.java
npalaska: @shogun656 good idea

src/main/java/io/auklet/agent/AukletUnhandledException.java
npalaska: @shogun656 will do

src/main/java/io/auklet/agent/AukletUnhandledException.java
npalaska: @shogun656 yes I was thinking to use `java.util.logging` instead of just stdout after doing some cleanups

src/main/java/io/auklet/agent/Device.java
npalaska: @rjenkinsjr We actually dont need to instantiate the class to represent the device because there is not gonna be multiple device instances once the app starts running

README.md
npalaska: @rjenkinsjr I have compiled it on java 9, that's the reason I didn't say it will work with java 8, I will be doing some testing to see it works with java 8 but if not I will have to recompile the agent with java 8

.idea/modules/agent/agent.iml
shogun656: ^

gradlew
shogun656: ^

src/main/java/io/auklet/agent/Auklet.java
shogun656: Do you want to do this in a different PR?

src/main/java/io/auklet/agent/Device.java
shogun656: ^

gradlew
rjenkinsjr: @shogun656 this file wasn't written by @npalaska, it's standard Gradle boilerplate and should not be modified by us.

src/main/java/io/auklet/agent/Auklet.java
npalaska: @shogun656 Yes I will open a new PR for it

README.md
rjenkinsjr: Were you able to test this yet?

src/main/java/io/auklet/agent/Auklet.java
rjenkinsjr: ^

Also, this class contains only `static` members, so should the class itself be `static`? If so, give it a `private` no-arg, empty-body constructor so it cannot be constructed.

src/main/java/io/auklet/agent/Device.java
```diff
@@ -0,0 +1,172 @@
...
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
+import java.io.*;
+
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.util.Scanner;
+
+public final class Device {
```
rjenkinsjr: This class contains only `static` members, so should the class itself be `static`? If so, give it a `private` no-arg, empty-body constructor so it cannot be constructed.

src/main/java/io/auklet/agent/MQTT.java
```diff
@@ -0,0 +1,131 @@
...
+import javax.net.ssl.TrustManagerFactory;
+import java.io.BufferedInputStream;
+import java.io.FileInputStream;
+import java.nio.charset.StandardCharsets;
+import java.security.KeyStore;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.util.Scanner;
+
+public final class MQTT {
```
rjenkinsjr: This class contains only `static` members, so should the class itself be `static`? If so, give it a `private` no-arg, empty-body constructor so it cannot be constructed.

src/main/java/io/auklet/agent/Messages.java
```diff
@@ -0,0 +1,35 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.json.simple.JSONObject;
+import org.msgpack.jackson.dataformat.MessagePackFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+public final class Messages {
```
rjenkinsjr: This class contains only `static` members, so should the class itself be `static`? If so, give it a `private` no-arg, empty-body constructor so it cannot be constructed.

src/main/java/io/auklet/agent/util.java
```diff
@@ -0,0 +1,95 @@
...
+import java.net.InetAddress;
+import java.net.NetworkInterface;
+import java.net.SocketException;
+import java.net.UnknownHostException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.HashMap;
+import java.util.Map;
+
+public final class util {
```
rjenkinsjr: This class contains only `static` members, so should the class itself be `static`? If so, give it a `private` no-arg, empty-body constructor so it cannot be constructed.

Also, this class name needs to start with a capital letter; it is not proper Java style to have classes that start with anything except uppercase letters. (You'll need to rename the file itself too.)

README.md
npalaska: @rjenkinsjr Yes I did some testing however since I compiled the agent with java 9 it was only supporting the application that were compiled in java 9. However adding ```sourceCompatibility = '1.8'
targetCompatibility = '1.8'``` to build.gradle does fix this compatibility issue and can now support java 8 as well, I will add that commit to this PR

src/main/java/io/auklet/agent/Device.java
npalaska: Yeah I think I can change this class as static, it shouldnt be a problem

src/main/java/io/auklet/agent/Auklet.java
npalaska: @rjenkinsjr yes but I think Java has no way of making a top-level class static we can only make nested classes as static. Making an outer class final would be enough since it can not be extended and so making it look like a static

src/main/java/io/auklet/agent/Messages.java
npalaska: @rjenkinsjr I dont think we can make outer classes static, making them only final should serve the purpose I think

README.md
```diff
@@ -25,8 +25,22 @@ public class demo {
        ...
    }
 }
-```
 
+// or if you have your own implementation of JVM shutdown hook we recommend the following 
+
+public class demo {
```
rjenkinsjr: Please use capital letters for class names.

PR #18, opened 2018-10-12
Do not fetch loopback network interface to calculate mac address hash 

npalaska:


src/main/java/io/auklet/agent/Util.java
```diff
@@ -16,12 +15,22 @@ private Util(){ }
     protected static String getMacAddressHash() {
         InetAddress ip;
         String machash = "";
+        NetworkInterface networkinterface = null;
         try {
 
-            ip = InetAddress.getLocalHost();
-            NetworkInterface network = NetworkInterface.getByInetAddress(ip);
+            System.out.println("Host addr: " + InetAddress.getLocalHost().getHostAddress());  // often returns "127.0.0.1"
```
bleib1dj: Shouldn't have any logging statements like this in the agent.

src/main/java/io/auklet/agent/Util.java
```diff
@@ -16,12 +15,22 @@ private Util(){ }
...
+            Enumeration<NetworkInterface> n = NetworkInterface.getNetworkInterfaces();
+            for (; n.hasMoreElements();)
+            {
+                NetworkInterface e = n.nextElement();
+                System.out.println("Interface: " + e.getName());
```
bleib1dj: This system out should be removed

npalaska: @bleib1dj indeed

PR #23, opened 2018-10-18
Add more error handling

npalaska:
This PR tries to add more robust error handling in java agent. 
If any of the following fails we do not set the unhandled exception handler 
    - We can't get the CA cert 
    - Can't register a device
    - Could not connect to MQTT client

### DIFF
--- a/src/main/java/io/auklet/agent/Device.java
+++ b/src/main/java/io/auklet/agent/Device.java
@@ -37,7 +37,7 @@ public final class Device {
     private static String client_password;
     private static String organization;
 
-    public static void register_device(String folderPath){
+    public static boolean register_device(String folderPath){
 
         JSONParser parser = new JSONParser();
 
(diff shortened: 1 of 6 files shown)


src/main/java/io/auklet/agent/Device.java
```diff
@@ -75,23 +79,22 @@ private static JSONObject create_device(){
...
-            String text = null;
-            try (Scanner scanner = new Scanner(response.getEntity().getContent(), StandardCharsets.UTF_8.name())) {
-                text = scanner.useDelimiter("\\A").next();
+            if(response.getStatusLine().getStatusCode() == 201) {
+                String text = null;
+                try (Scanner scanner = new Scanner(response.getEntity().getContent(), StandardCharsets.UTF_8.name())) {
+                    text = scanner.useDelimiter("\\A").next();
+                }
+                JSONParser parser = new JSONParser();
+                JSONObject myResponse = (JSONObject) parser.parse(text);
```
shogun656: I am sure this will error out if text is null. you want to put this in the try catch statement

src/main/java/io/auklet/agent/MQTT.java
```diff
@@ -115,14 +117,19 @@ private static JSONObject getbroker(){
...
+                String text = null;
+                try (Scanner scanner = new Scanner(response.getEntity().getContent(), StandardCharsets.UTF_8.name())) {
+                    text = scanner.useDelimiter("\\A").next();
+                }
 
-            JSONParser parser = new JSONParser();
-            JSONObject brokers = (JSONObject) parser.parse(text);
-            return brokers;
+                JSONParser parser = new JSONParser();
+                JSONObject brokers = (JSONObject) parser.parse(text);
```
shogun656: same with here

src/main/java/io/auklet/agent/Device.java
npalaska: good point

PR #29, opened 2018-10-29
APM-1649: Add slf4j logging in Auklet agent code

npalaska:


src/main/java/io/auklet/agent/Auklet.java
```diff
@@ -3,17 +3,20 @@
...
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public final class Auklet {
 
     static protected String AppId;
     static protected String ApiKey;
     static protected MqttClient client;
+    static protected Logger logger = LoggerFactory.getLogger("Auklet");
```
bleib1dj: @npalaska I’m don’t think you want to have this be a static property on the Auklet class. Instead I believe you should be initializing the logger in each file independently. That way you don’t run the risk of circular dependency and avoid bring in the entire Auklet class for situations where you just need the logger.

npalaska: @bleib1dj interesting point but I did a little research and it won't cause the circular dependency since the logger field is declared static. Static fields are class variables and are shared amongst all instances of that class. Circular dependency occurs when class A instance depends on class B instance and vice versa. However, we also can not create an instance of Auklet class since its declared final. Also, classloader usually loads the class only once in memory even if it is used in multiple time in the application and static fields of those classes are stored in permanent memory (heap in Hotspot JVM).

npalaska:
@shogun656 do you think we need to use the slf4j to log4j bridge in the agent? https://www.slf4j.org/manual.html

npalaska:
@bleib1dj do you have any comments for this PR now, if not I think it should be okay to get merged.

src/main/java/io/auklet/agent/Auklet.java
rjenkinsjr: The issue I see with the current implementation in this PR is not that the alternative (initializing one logger per class) would result in circular dependencies or excessive resource consumption, but rather that the alternative is the standard way in which Java applications/libraries log information. The most important reason for this is so that it is clear in the log message which class did the logging, which makes debugging much easier. As this PR stands now, I have no way of seeing which classes are logging which messages because they all share a logger. It's counterintuitive and would make debugging harder for us, and for end users, down the road. I don't really see a reason to deviate from standard industry practice in this regard unless we are so concerned about resource utilization that we want to optimize every single usage of the JVM heap; I can see in principle why we would want to do this, but honestly that feels like an over-optimization to me.

The standard way of initializing loggers in SLF4J is to use code such as `Logger logger = LoggerFactory.getLogger(HelloWorld.class);`, which is a convenience method that will initialize a logger using the FQCN of `HelloWorld` as its name. You _do_ want this to be a static variable per-class, even for non-singleton classes, because creating a logger per-object could become unnecessarily resource intensive and also rarely adds value to the overall effort of logging messages.

build.gradle
```diff
@@ -33,6 +33,7 @@ dependencies {
     compile "com.googlecode.json-simple:json-simple:1.1.1"
     compile "org.msgpack:jackson-dataformat-msgpack:0.8.16"
     compile "org.eclipse.paho:org.eclipse.paho.client.mqttv3:1.2.0"
+    compile "org.slf4j:slf4j-simple:1.7.25"
```
rjenkinsjr: This is not the correct dependency to use. You need to depend upon `slf4j-api` so as to not impose the `simple` logging implementation on end users.

If you want to use `simple` as an implementation for testing purposes, you need to declare that dependency using a different scope such as `test`. I'm not as familiar with Gradle scopes as I am with Maven scopes so I'm not sure what the right choice is here.

src/main/java/io/auklet/agent/Auklet.java
npalaska: @rjenkinsjr yes I agree, having each class implement its own logger with the respective class name would make it easier for us to understand the logged information. However, the reason I did this was that we don't have too many classes and object definitions and our project structure is quite simpler but yes we shouldn't digress from the standard practices.

build.gradle
npalaska: I see

npalaska: @rjenkinsjr I am getting this error `SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder"` which seems like a pretty known bug in using `slf4j-api`. The documentation here https://www.slf4j.org/manual.html also suggest we use both `slf4j-api` with `slf4j-simple`

src/main/java/io/auklet/agent/AukletExceptionHandler.java
```diff
@@ -44,10 +47,9 @@ protected static AukletExceptionHandler setup() {
 
     protected static synchronized void sendEvent(Throwable thrown) {
         List<Object> list = new ArrayList<>();
-        System.err.print("Exception in thread \"" + Thread.currentThread().getName() + "\" ");
-        thrown.printStackTrace(System.err);
+        logger.info("Exception in application thread \"" + Thread.currentThread().getName() + "\" ");
 
-        System.out.println("Exception message from app  " + thrown.getMessage());
+        logger.info("Exception message from application: " + thrown.getMessage());
```
rjenkinsjr: You probably want to use a finer logging level than `INFO` here. `DEBUG` is probably a good fit. I think this will create too much noise.

src/main/java/io/auklet/agent/SystemMetrics.java
```diff
@@ -25,6 +30,7 @@ protected static void initSystemMetrics(){
         }catch (Exception e) {
             // If Exception occurs, it mean we can not use com.sun.management package and
             // we fall back without recording mem usage
+            logger.warn("Underlying JVM does not support sun framework implementation");
```
rjenkinsjr: I'm not sure this will work the way you're expecting. My hunch is that if you attempt to run the Java agent on a JVM that doesn't have these classes, you won't even get this far - I expect you'd run into a classloading exception for this class (`SystemMetrics`) because it wouldn't be able to resolve this reference.

Please test this on a non-conformant JVM implementation and let me know what happens.

src/main/java/io/auklet/agent/Util.java
```diff
@@ -28,6 +32,7 @@ protected static String getMacAddressHash() {
                     break;
                 }
             }
+            logger.info("Network Interface: " + networkinterface);
```
rjenkinsjr: This is probably too noisy, might want to drop this to `DEBUG` instead.

src/main/java/io/auklet/agent/Util.java
```diff
@@ -72,11 +77,11 @@ protected static String createCustomFolder(String sysProperty) {
         String path = System.getProperty(sysProperty) + File.separator + "aukletFiles";
         File newfile = new File(path);
         if (newfile.exists()){
-            System.out.println("folder already exists");
+            logger.info("Folder already exists");
```
rjenkinsjr: This is probably too noisy, might want to drop this to `DEBUG` instead. Same for the statements below this one.

src/main/java/io/auklet/agent/Device.java
```diff
@@ -93,14 +99,13 @@ private static JSONObject create_device(){
             }
 
             else {
-                System.out.println("could not create a device and status code is: " +
+                logger.error("could not create a device and status code is: " +
```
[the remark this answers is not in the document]
npalaska: The response code that we would receive will contain `client__password`, `client_username` etc if we get it right. However, I am logging only when we don't get response code 201, and so I am just logging the response code which should give us a good idea about what went wrong

src/main/java/io/auklet/agent/AukletExceptionHandler.java
```diff
@@ -44,10 +47,9 @@ protected static AukletExceptionHandler setup() {
 
     protected static synchronized void sendEvent(Throwable thrown) {
         List<Object> list = new ArrayList<>();
-        System.err.print("Exception in thread \"" + Thread.currentThread().getName() + "\" ");
-        thrown.printStackTrace(System.err);
+        logger.info("Exception in application thread \"" + Thread.currentThread().getName() + "\" ");
```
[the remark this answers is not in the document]
npalaska: The `uncaughtException` method catches the exception thrown in any thread that is running at a given time. `uncaughtException` catches the thread and throwable exception and once it is caught we send an event. However, we do not send any information about the particular thread that caused the exception, we only send the exception. Therefore I am using only throwable as a parameter for `sendEvent` method

src/main/java/io/auklet/agent/AukletExceptionHandler.java
npalaska: @rjenkinsjr the reason I used `info` and not `DEBUG` is that the logged information does not concern the agent but these are the things that happened in the java application that is running the agent. So these are not the debugging exceptions from the agent itself.

src/main/java/io/auklet/agent/SystemMetrics.java
npalaska: @rjenkinsjr I don't have any import statements for `sun` classes and as java loads the classes on demand, it will try to load this class only when the `getSystemMetrics` class is called. many of the class loading exceptions such as `classDefNotFound` or `classCastException` are inherited from class `Exception` therefore my guess was that it would get caught. 
However, I see your concern and I will be doing more agent testing on non-oracle based JVMs later.

rjenkinsjr: A proper check for this would be:

```
boolean hasSunBeanClass = true
try {
  Class.forName("com.sun.management.OperatingSystemMXBean")
} catch (ClassNotFoundException e) {
  hasSunBeanClass = false
  // Class is missing, do whatever fallback behavior is needed.
  // This exception is expected in this context, so not necessary to log it.
}
if (hasSunBeanClass) {
  // do stuff with the class
}
```

This assumes that my hunch is wrong and that using the class via FQCN in the code rather than via an `import` statement is legal at runtime if it is missing. If my hunch is right, you'll end up with a `NoClassDefFoundError` when you attempt to use anything in _this_ class (not the class that is missing). You'll just have to try it out to see if that's the case. If that does happen, note that `NoClassDefFoundError` is a subclass of `Error`, not `Exception`, so a generic `catch(Exception e)` won't handle it.

Please test on a non-conformant JVM (e.g. IBM) and let's see how it behaves so we can make the necessary changes (if any) to this PR.

src/main/java/io/auklet/agent/AukletExceptionHandler.java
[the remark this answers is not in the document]
rjenkinsjr: I don't think I explained my point well. `Thread.currentThread().getName()` is doomed to return the wrong thread name associated with the given exception in virtually all cases, because the _current thread_ is most definitely not the thread where the exception occurred. If that were true, `uncaughtException` would not need a `thread` argument at all.

If the intent is to log to SLF4J for debugging purposes where the exception occurred, then `Thread.currentThread().getName()` is not correct.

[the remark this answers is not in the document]
npalaska: Ohh now I see what you mean @rjenkinsjr I think you are right

src/main/java/io/auklet/agent/AukletExceptionHandler.java
npalaska: @rjenkinsjr that was just me doing some testing to confirm whether the exception we caught is the same we are sending to the backend. I can probably take this out now.

src/main/java/io/auklet/agent/Device.java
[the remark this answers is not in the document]
bleib1dj: @rjenkinsjr DRF helps us out of the box on this front. For everything other than a 404 I would agree with @npalaska, we should be able to interpret the issue from the status code and the body won't help us much more than if you were to go google what the status code stands for. 

In the case of a 404 though, an endpoint will return back what fields failed the serializer validation and may also provide insights on how to fix it. 

In regards to the response containing the `client_password` and `client_username` that would only happen if we logged off the 201 response @npalaska. Since you are not entering this block when you receive a 201 you shouldn't need to worry about that. 

So all that being said I'd take the following approach. For all non-2xx/3xx log an error with the status code and the body of the response. For any 2xx or 3xx responses don't log out any info.

src/main/java/io/auklet/agent/SystemMetrics.java
npalaska: @rjenkinsjr I compiled and executed java agent on IBM JVM and interestingly I did not get any error. I was using Java 8 SDK from IBM. When I dig down into the documentation I found out they have new `MXbean` implementation which extends from `com.sun.management` https://www.ibm.com/support/knowledgecenter/en/SSYKE2_8.0.0/com.ibm.java.api.80.doc/com.ibm.lang.management/index.html and maybe I am guessing that's why `com.sun.management` package got imported even with IBM JVM platform.

npalaska: So what I would say @rjenkinsjr is that I will open a new PR if needed any implementation changes for any JVM that cannot import `com.sun.` packages (which apparently IBM can). So we don't mix up things and digress from the actual PR

rjenkinsjr: Probably a good idea since this isn't related to logging. In that case, please remove this logic from this PR.

npalaska: @rjenkinsjr I did not add this logic as a part of this PR. I think it's better to keep the logic as it is for now since we do not know for sure whether it will raise an exception or gonna error out on any JVM that does not support importing `com.sun` packages. Once I do the testing with more non-conformant JVMs I will do the necessary modifications if needed.

src/main/java/io/auklet/agent/Device.java
```diff
@@ -82,18 +82,19 @@ private static JSONObject create_device() {
             request.setEntity(params);
             HttpResponse response = httpClient.execute(request);
 
+            String text;
+            try (Scanner scanner = new Scanner(response.getEntity().getContent(), StandardCharsets.UTF_8.name())) {
```
rjenkinsjr: This code is duplicated elsewhere in this file. Can we extract it into a utility method?

src/main/java/io/auklet/agent/Device.java
```diff
@@ -82,18 +82,19 @@ private static JSONObject create_device() {
...
-                } catch (Exception e) {
-                    logger.error("Exception during reading contents of create device endpoint: " + e.getMessage());
-                    return null;
-                }
                 return new JSONObject(text);
             } else {
-                logger.error("could not create a device and status code is: " +
-                        response.getStatusLine().getStatusCode());
+                logger.error("could not create a device and status code is: " + response.getStatusLine());
+                logger.error("could not create a device and response is: " + text);
```
rjenkinsjr: Let's collapse these two log statements as follows:

`logger.error(String.format("Error while creating device: %s: %s", response.getStatusLine(), text))`

@npalaska I assume `getStatusLine` contains the status code as well? I see you dropped `getStatusCode` in favor of `getStatusLine`.

src/main/java/io/auklet/agent/Device.java
```diff
@@ -162,21 +163,26 @@ public static boolean get_Certs(String folderPath) {
 
                 HttpGet request = new HttpGet(con.getURL().toURI());
                 HttpResponse response = httpClient.execute(request);
-                if (response.getStatusLine().getStatusCode() == 200) {
-                    InputStream ca = response.getEntity().getContent();
-                    String text;
-                    try (Scanner scanner = new Scanner(ca, StandardCharsets.UTF_8.name())) {
-                        text = scanner.useDelimiter("\\A").next();
-                    }
+                InputStream ca = response.getEntity().getContent();
```
rjenkinsjr: Don't forget to close this stream, while taking exception handling into account (e.g. use `try/finally` or try-with-resources).

src/main/java/io/auklet/agent/Device.java
```diff
@@ -162,21 +163,26 @@ public static boolean get_Certs(String folderPath) {
...
+
+                if (response.getStatusLine().getStatusCode() == 200) {
                     BufferedWriter writer = new BufferedWriter(new FileWriter(folderPath + "/CA"));
                     writer.write(text);
                     writer.close();
                     logger.info("CA File is created!");
                 } else{
-                    logger.info("Get cert response code: " + response.getStatusLine().getStatusCode());
+                    logger.error("Get cert response code: " + response.getStatusLine());
+                    logger.error("Get cert response body: " + text);
```
rjenkinsjr: Collapse these two log statements in a similar fashion to what I proposed above.

src/main/java/io/auklet/agent/MQTT.java
```diff
@@ -106,18 +106,20 @@ private static JSONObject getbroker() {
...
-                } catch (Exception e) {
-                    logger.error("Exception occurred during reading brokers info: " + e.getMessage());
-                    return null;
-                }
                 return new JSONObject(text);
             }
             else {
-                logger.info("get broker response code: "+ response.getStatusLine().getStatusCode());
+                logger.info("get broker response code: "+ response.getStatusLine());
+                logger.info("get broker response body: "+ text);
```
rjenkinsjr: Collapse these two log statements in a similar fashion to what I proposed above.

src/main/java/io/auklet/agent/Device.java
```diff
@@ -162,21 +163,26 @@ public static boolean get_Certs(String folderPath) {
...
                     if(file.delete()){
-                        logger.info("CA file deleted");
+                        logger.error("CA file deleted");
```
rjenkinsjr: I'm not understanding the logic here. If the file is deleted successfully, we announce it as such, but as an error?

src/main/java/io/auklet/agent/AukletExceptionHandler.java
```diff
@@ -61,11 +60,10 @@ protected static synchronized void sendEvent(Throwable thrown) {
             message.setQos(2);
             Auklet.client.publish("java/events/" + Device.getOrganization() + "/" +
                     Device.getClient_Username(), message);
-            System.out.println("Message published");
+            logger.info("Message published");
```
rjenkinsjr: To make this log message more useful during debugging, is there any non-sensitive unique identifier that we could log?

src/main/java/io/auklet/agent/SystemMetrics.java
rjenkinsjr: @npalaska my bad, thanks for clarifying. Somehow I got the impression that this logic was added in this PR.

src/main/java/io/auklet/agent/Device.java
npalaska: @rjenkinsjr yes I think I can create a utility method for this

npalaska: @rjenkinsjr I just checked the PR opened by @shogun656, he has taken care of creating a utility method for `httpGet` and `httpPost` logic.

src/main/java/io/auklet/agent/AukletExceptionHandler.java
npalaska: I dont know if there is anything unique to the MQTT messages unless we set an id for each message, probably we can log whether the published message is duplicate or not https://www.eclipse.org/paho/files/javadoc/org/eclipse/paho/client/mqttv3/MqttMessage.html

src/main/java/io/auklet/agent/AukletExceptionHandler.java
```diff
@@ -60,10 +61,10 @@ protected static synchronized void sendEvent(Throwable thrown) {
             message.setQos(2);
             Auklet.client.publish("java/events/" + Device.getOrganization() + "/" +
                     Device.getClient_Username(), message);
-            logger.info("Message published");
+            logger.info(String.format("Duplicate message published: %s", message.isDuplicate()));
```
rjenkinsjr: If `message.isDuplicate()` returns a `Boolean`, then `%s` needs to be changed to `%b`. See https://docs.oracle.com/javase/8/docs/api/java/util/Formatter.html for details.

src/main/java/io/auklet/agent/Device.java
```diff
@@ -44,7 +44,7 @@ public static boolean register_device(String folderPath) {
         try {
             Path fileLocation = Paths.get(folderPath + filename);
             byte[] data = Files.readAllBytes(fileLocation);
-            logger.info("AukletAuth file content length: " + data.length);
+            logger.info(String.format("AukletAuth file content length: %s", data.length));
```
rjenkinsjr: `data.length` is an integer, so `%s` needs to be changed to `%d`. See https://docs.oracle.com/javase/8/docs/api/java/util/Formatter.html for details.

npalaska:
@rjenkinsjr I agree that end users decide which logging implementation and logging configuration to use and we shouldn't dictate it. However, these logs are for coming from the agents and not the application itself, these logs do not concern the end users it's for us to understand what might have gone wrong. The reason I logged `Throwable.toString()` is that I wasn't sure if logging stack traces from java agent is something overkilling it, but when I did some research it seems like logging just the `Throwable` is pretty common practice.

src/main/java/io/auklet/agent/Auklet.java
```diff
@@ -40,7 +40,7 @@ private static void setup(String appId, String apiKey, boolean handleShutDown){
                             logger.info("Auklet agent shutting down");
                             Auklet.shutdown();
                         } catch (Exception e) {
-                            logger.error(String.format("error while shutting down Auklet agent %s", e.toString()));
+                            logger.error("error while shutting down Auklet agent ", e);
```
rjenkinsjr: Capitalize `error` as `Error`.

src/main/java/io/auklet/agent/Device.java
```diff
@@ -60,7 +60,7 @@ public static boolean register_device(String folderPath) {
             } else return false;
 
         } catch (Exception e) {
-            logger.error(String.format("Exception while device registration: %s", e.toString()));
+            logger.error("Exception while device registration ", e);
```
rjenkinsjr: Say `during` instead of `while`.

src/main/java/io/auklet/agent/Device.java
```diff
@@ -44,7 +44,7 @@ public static boolean register_device(String folderPath) {
         try {
             Path fileLocation = Paths.get(folderPath + filename);
             byte[] data = Files.readAllBytes(fileLocation);
-            logger.info(String.format("AukletAuth file content length: %s", data.length));
+            logger.info(String.format("AukletAuth file content length: %d", data.length));
```
rjenkinsjr: Here you have it written as `AukletAuth file` but elsewhere in this PR you have it written as `Auklet Auth file`. For consistency let's say `Auklet auth file`.

src/main/java/io/auklet/agent/Device.java
```diff
@@ -173,10 +173,10 @@ public static boolean get_Certs(String folderPath) {
                     }
                 }
             } else {
-                logger.info("CA File already exists.");
+                logger.info("CA File already exists");
```
rjenkinsjr: Do not capitalize `File`.

src/main/java/io/auklet/agent/MQTT.java
```diff
@@ -89,7 +89,7 @@ private static SSLSocketFactory getSocketFactory (String caFilePath) {
 
             return context.getSocketFactory();
         } catch (Exception e) {
-            logger.error(String.format("Error while setting up socket factory: %s", e.toString()));
+            logger.error("Error while setting up socket factory ", e);
```
rjenkinsjr: Let's say `setting up MQTT socket factory` to eliminate any future possible ambiguity.

src/main/java/io/auklet/agent/Messages.java
```diff
@@ -33,7 +33,7 @@ private Messages(){ }
             bytes = objectMapper.writeValueAsBytes(map);
 
         } catch (JsonProcessingException e) {
-            logger.error(String.format("Exception while processing json in creating messagePack: %s", e.toString()));
+            logger.error("Exception while processing json in creating messagePack ", e);
```
rjenkinsjr: Let's say `Exception while converting JSON to msgpack`.

src/main/java/io/auklet/agent/Util.java
```diff
@@ -61,7 +61,7 @@ protected static String getIpAddress() {
 
             ipAddr = in.readLine(); //you get the IP as a String
         } catch (IOException e) {
-            logger.error(String.format("Error while fetching the ip address: %s", e.toString()));
+            logger.error("Error while fetching the ip address ", e);
```
rjenkinsjr: Capitalize `ip` as `IP`.

src/main/java/io/auklet/agent/Util.java
```diff
@@ -47,7 +47,7 @@ protected static String getMacAddressHash() {
             machash = Hex.encodeHexString(macHashByte);
 
         } catch (SocketException | NoSuchAlgorithmException | UnsupportedEncodingException e) {
-            logger.error(String.format("Error while computing the mac address hash: %s", e.toString()));
+            logger.error("Error while computing the mac address hash ", e);
```
rjenkinsjr: Capitalize `mac` as `MAC`.

src/main/java/io/auklet/agent/Util.java
```diff
@@ -81,13 +81,12 @@ protected static String createCustomFolder(String sysProperty) {
...
 
-    protected static String readContents(String methodName, HttpResponse response) {
+    protected static String readContents(HttpResponse response) {
         String text;
         try (Scanner scanner = new Scanner(response.getEntity().getContent(), StandardCharsets.UTF_8.name())) {
             text = scanner.useDelimiter("\\A").next();
         } catch (Exception e) {
-            logger.error(String.format("Exception while parsing the contents of response using scanner in %s: %s",
-                    methodName, e.toString()));
+            logger.error("Exception while parsing the contents of response using scanner in ", e);
```
rjenkinsjr: Let's say `Exception while parsing HTTP response body` for brevity. If someone wants to know that we're using `Scanner` they'll either see it in the stack trace or they can dig into the source code.

src/main/java/io/auklet/agent/Auklet.java
```diff
@@ -92,7 +92,7 @@ public static void shutdown() {
         try {
             client.close();
         } catch (MqttException e) {
-            logger.error("Error while closing mqtt client ", e);
+            logger.error("Error while closing mqtt client", e);
```
rjenkinsjr: Capitalize `mqtt`.

src/main/java/io/auklet/agent/AukletExceptionHandler.java
```diff
@@ -64,7 +64,7 @@ protected static synchronized void sendEvent(Throwable thrown) {
             logger.info(String.format("Duplicate message published: %b", message.isDuplicate()));
 
         } catch (MqttException | NullPointerException e) {
-            logger.error("Error while publishing the mqtt message ", e);
+            logger.error("Error while publishing the mqtt message", e);
```
rjenkinsjr: Capitalize `mqtt`.

src/main/java/io/auklet/agent/Device.java
```diff
@@ -40,20 +44,23 @@ public static boolean register_device(String folderPath) {
...
             byte[] data = Files.readAllBytes(fileLocation);
+            logger.info(String.format("Auklet auth file content length: %d", data.length));
+
             Cipher cipher = Cipher.getInstance("AES");
             cipher.init(Cipher.DECRYPT_MODE, aesKey);
             String decrypted = new String(cipher.doFinal(data));
             setCreds(new JSONObject(decrypted));
 
         } catch (FileNotFoundException | NoSuchFileException e) {
+            logger.info("Creating a new AukletAuth file");
```
rjenkinsjr: Should be written as `Auklet auth file`.

src/main/java/io/auklet/agent/Device.java
```diff
@@ -156,30 +156,27 @@ public static boolean get_Certs(String folderPath) {
...
+
+                if (response.getStatusLine().getStatusCode() == 200 && contents != null) {
                     BufferedWriter writer = new BufferedWriter(new FileWriter(folderPath + "/CA"));
-                    writer.write(text);
+                    writer.write(contents);
                     writer.close();
-                    System.out.println("CA File is created!");
-                } else {
-                    System.out.println("Get cert response code: " + response.getStatusLine().getStatusCode());
+                    logger.info("CA File is created!");
```
rjenkinsjr: Should be written as `CA file`.

src/main/java/io/auklet/agent/Device.java
```diff
@@ -75,23 +82,16 @@ private static JSONObject create_device() {
...
-                }
-                return new JSONObject(text);
+            String contents = Util.readContents(response);
+
+            if(response.getStatusLine().getStatusCode() == 201 && contents != null) {
+                return new JSONObject(contents);
             } else {
-                System.out.println("could not create a device and status code is: " +
-                        response.getStatusLine().getStatusCode());
+                logger.error(String.format("Error while creating device: %s: %s", response.getStatusLine(), contents));
```
rjenkinsjr: This comment applies to 10 areas in this PR where you use `String.format`.

I forgot when I suggested `String.format` to you that SLF4J has its own parameterized logging feature. You can write this logging statement as follows to achieve the same effect: `logger.error("Error while creating device: {}: {}", response.getStatusLine(), contents);`.

The substitution marker is `{}` regardless of the type of object being passed as an argument. SLF4J calls `toString()` on arguments where applicable, so anywhere you pass other primitives (e.g. integers or booleans) you can just pass those values verbatim.

Please update the 10 places in this PR where you use `String.format` to use this parameterization technique.

PR #32, opened 2018-11-06
APM-1605: Device Configuration Integration

shogun656:
This PR adds the config options from the API. This PR also converts the MQTT client to an Async Client.

This PR is blocked by:
https://github.com/aukletio/Auklet-Agent-Java/pull/29
https://github.com/aukletio/Auklet-Agent-Java/pull/31

### DIFF
--- a/src/main/java/io/auklet/agent/DataRetention.java
+++ b/src/main/java/io/auklet/agent/DataRetention.java
@@ -7,18 +7,26 @@ import org.json.JSONObject;
 import java.io.File;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.util.Calendar;
+import java.util.Timer;
+import java.util.TimerTask;
 
 public class DataRetention {
 
     final static private Long MEGABYTES_TO_BYTES = 1000000L;
     final static private Long SECONDS_TO_MILLISECONDS = 1000L;
 
+    static private boolean freshStart = true;
+    static private boolean resetData = false;
+
     static private Long emissionPeriod = 60000L;
     static private Long storageLimit;
     static private Long cellularDataLimit;
     static private int cellPlanDate;
-    static private Long dataSent;
+
     static private String usageFile;
+    static private Long dataSent;
+    static private int hours;
 
     private DataRetention(){ }
 
@@ -30,6 +38,11 @@ public class DataRetention {
 
         ObjectMapper mapper = new ObjectMapper();
         dataSent = mapper.readValue(new File("json_file"), JSONObject.class).getLong("usage");
+
+        if (freshStart) {
+            checkDate();
+            freshStart = false;
+        }
     }
 
     public static void updateDataSent(int dataSize) {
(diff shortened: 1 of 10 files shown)


src/main/java/io/auklet/agent/DataRetention.java
```diff
@@ -0,0 +1,110 @@
...
+        return (dataSent + dataSize <= cellularDataLimit);
+    }
+
+    public static void setUsageFile(String file) throws IOException {
+        usageFile = file;
+        File usageFile = new File(file);
+        if (usageFile.createNewFile()) {
+            writeToUsageFile(0L);
+        }
+    }
```
bleib1dj: @shogun656 know we do it this way in the other agents but as I think about it more I believe this may cause issues. Especially if we're doing it async. My fear is that by having n number of threads attempting to write to the file could cause the file to become corrupted. It could also cause exceptions to be thrown that cause problems for the agent or the monitored application.

We should look into using a mutex or some alternative method for improving the process in which we access the file. Found a few relevant articles / docs:

* https://stackoverflow.com/questions/28759584/reading-and-writing-in-files-with-thread-and-semaphore
* https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/ReadWriteLock.html
* https://docs.oracle.com/javase/tutorial/essential/concurrency/locksync.html
* https://stackoverflow.com/questions/15183842/how-to-create-interprocess-file-access-mutex-in-java

src/main/java/io/auklet/agent/Device.java
```diff
@@ -184,4 +174,52 @@ public static boolean get_Certs(String folderPath) {
...
+                }
+                JSONObject conf = new JSONObject(text).getJSONObject("config");
+                DataRetention.initDataRetention(conf);
+
+                FileWriter writer = new FileWriter(limits);
+                writer.write(conf.toString());
+                writer.close();
+                System.out.println("Config File was stored");
+            } else {
+                System.out.println("Get config response code: " + response.getStatusLine().getStatusCode());
```
bleib1dj: Shouldn't we be using the logging interface rather than `System.out.println`? (Same question applies to all usages of `System.out` in this PR)

src/main/java/io/auklet/agent/DataRetention.java
kdsch: I have no place in a conversation about Java implementations, but nevertheless I'd like to mention message passing and shared-nothing concurrency as an alternative. The main advantages over mutexes are composability and scalability. If the size of the critical section can be expected to increase, beware of mutexes. Otherwise, mutexes can be fast and maintainable.

The C/C++ client, for example, has a single goroutine that reads and writes to the usage file. There is no shared access. The goroutine reads the stream (or queue) of outbound messages. For each message, its size is calculated and the file is updated. Then the message is relayed to the next stage, whoever that is (perhaps an MQTT client).

src/main/java/io/auklet/agent/Device.java
shogun656: I was waiting for the logging PR to be pulled in before I switch to logging. That's why I put it as a blocker for this PR. Once that's pulled in I will switch all the prints to logs

src/main/java/io/auklet/agent/DataRetention.java
shogun656: Hmm thanks for the information guys. I settled with adding `synchronized` to `writeToUsageFile` that way only one thread can access that function at a time. There shouldn't be many cases of collision so I think this should be sufficient and it is only a one line change.

rjenkinsjr: @shogun656 I recommend you use a `synchronized` block on a `private static` lock object rather than the method modifier. Using the method modifier opens up the possibility of a deadlock-based DoS attack (an attacker could very easily get a reference to the `Class<?>` object and then `synchronize` on that, causing `writeToUsageFile` to lock forever).

See https://stackoverflow.com/a/20907356 for an explanation/short example. In your case, the method and lock object would both be `static`. (Btw, the book "Effective Java" referenced in that SO link is *very* good, I would recommend anyone interested in writing robust Java code to invest in a copy.)

shogun656: @rjenkinsjr Do you think my latest commit is better. I am locking based on the `private static usageFile` which is the file name. It is not recommended to lock based on a non-final field because setField could change that field and another thread could have access to the synchronize block. Although, I think in our case this would not be a problem because if `usageFile` does change then the thread would be writing to a different file and should not affect any other synchronized block. I could also do (private) final lock dummy as talked about in this link https://stackoverflow.com/a/21462631/6709710 but I think that won't be needed

PR #33, opened 2018-11-15
APM-1668: Java USB Communication

shogun656:
This PR refactors the MQTTClient into an interface design so we can switch between the MQTTClient and SerialClient very easily. This PR also adds the SerialClient so we have the ability to send events to Serial Interface

### DIFF
--- a/src/main/java/io/auklet/agent/Auklet.java
+++ b/src/main/java/io/auklet/agent/Auklet.java
@@ -1,7 +1,8 @@
 package io.auklet.agent;
 
 import io.auklet.agent.broker.Client;
-import io.auklet.agent.broker.MQTT;
+import io.auklet.agent.broker.MQTTClient;
+import io.auklet.agent.broker.SerialClient;
 
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -14,22 +15,52 @@ public final class Auklet {
 
     private Auklet(){ }
 
-    private static void setup(String appId, String apiKey, boolean handleShutDown, String serialOut){
+    private static void setup(String appId, String apiKey, boolean handleShutDown, String serialOut) {
         ApiKey = apiKey;
         AppId = appId;
 
+        ScheduledExecutorService mqttThreadPool = createThreadPool();
+        String folderPath = createFolderPath();
+        SystemMetrics.initSystemMetrics();
+
+        client = createClient(folderPath, serialOut, mqttThreadPool);
+
+        if (client != null && client.isSetUp()) {
+            AukletExceptionHandler.setup();
+            if(handleShutDown)
+                setUpShutdownThread(mqttThreadPool);
+        }
+    }
+
+    public static void init(String appId, String apiKey) {
+        setup(appId, apiKey, true, "");
+    }
+
+    public static void init(String appId, String apiKey, boolean handleShutDown) {
+        setup(appId, apiKey, handleShutDown, "");
+    }
+
+    public static void init(String appId, String apiKey, boolean handleShutDown, String serialOut) {
+        setup(appId, apiKey, handleShutDown, serialOut);
+    }
+
+    public static void exception(Throwable thrown) {
+        AukletExceptionHandler.sendEvent(thrown);
+    }
+
+    private static ScheduledExecutorService createThreadPool() {
         /*
         Ref: https://github.com/eclipse/paho.mqtt.java/issues/402#issuecomment-424686340
          */
-        ScheduledExecutorService mqttThreadPool = Executors.newScheduledThreadPool(10,
+        return Executors.newScheduledThreadPool(10,
                 (Runnable r) -> {
                     Thread t = Executors.defaultThreadFactory().newThread(r);
                     t.setDaemon(true);
                     return t;
                 });
+    }
 
-        SystemMetrics.initSystemMetrics();
-
+    private static String createFolderPath() {
         String folderPath = Util.createCustomFolder("user.dir");
         if (folderPath == null){
             folderPath = Util.createCustomFolder("user.home");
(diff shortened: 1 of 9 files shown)


shogun656:
This PR is ready for a review

src/main/java/io/auklet/agent/broker/SerialClient.java
```diff
@@ -0,0 +1,70 @@
...
+            setUp = true;
+        } catch (NoSuchPortException | PortInUseException | IOException e) {
+            setUp = false;
+            logger.error(e.toString());
+        }
+    }
+
+    @Override
+    public boolean isSetUp() {
+        return setUp;
```
kdsch: Not sure if this is a valuable input to the PR (since I'm not adept at Java), but I have an observation. This value is basically encoding whether the constructor succeeded or failed, right? Would it make sense to store the exception value instead? Then clients of this class could access the error; logging it could happen on demand rather than supply. Furthermore the error value could be used to convey other errors as well, not just constructor failure.

shogun656: I see what you mean but what could be the use of storing the error. As of right now, we have Auklet exit if the client was not created which can be checked by this `isSetUp` function. If we have a function that instead returns the exception, what would be the use for Auklet? What can Auklet do with the error object besides knowing to exit?

rjenkinsjr: @shogun656 the problem with the way the code is written now, and I think this is what @kdsch is touching on, it creates the illusion that the calling code has a functioning client object, when in reality it does not.

Right now, if an error occurs, you're merely logging it (improperly, btw - please use the `Throwable`-aware logging methods offered by the SLF4J API instead of using `e.toString()`, as the latter almost always hides the stack trace). Calling code then has to use `isSetUp` to figure out if it's safe to continue, rather than relying on exceptions to communicate that fact. As it is now, I cannot simply rely on whether or not the client object is constructed to know if it's safe to use - I also have to check the `isSetUp` method, which doesn't really serve any other purpose.

You don't need to "store the error"; you can just let the exception in the constructor bubble up. That way, you'd be relying on the exception mechanisms in Java to signal to calling code that something is wrong and that it should react accordingly (by not attempting to do things with an object that it failed to construct, e.g. by continuing to check if the variable intended to hold that object is `null`). While this particular example is code that's internal to the agent and never accessed by end users, the same principles apply.

So, if you were to change this constructor to say...

```
    public SerialClient(String portName) throws NoSuchPortException, PortInUseException, IOException {
        comm = (SerialPort) CommPortIdentifier.getPortIdentifier(portName).open("AukletPort", 1000);
        stream = comm.getOutputStream();
    }
```

...you can then in the calling code say...

```
// Not sure if it's best to handle the exception in the createClient method or the setup method, but the idea is the same either way, so below is a pseudocode example.
try {
	// If not MQTT, then...
    Client client = new SerialClient(serialOut)
    // If we reach this point, we know by definition that the client is properly constructed,
    // so now we can do stuff with it.
    //
    // You could also store this in a null-initialized variable that is declared prior to the try block,
    // and then after the try/catch check to see if the variable is still null.
    // Java won't set the variable to non-null if an exception occurs during object construction.
} catch (NoSuchPortException | PortInUseException | IOException e) {
    // log appropriately
}
```

src/main/java/io/auklet/agent/broker/SerialClient.java
```diff
@@ -0,0 +1,70 @@
...
+            byte[] bytes = objectMapper.writeValueAsBytes(map);
+            stream.write(bytes);
+            logger.info("Event was sent");
+        } catch (IOException e) {
+            logger.error("Event was unable to be sent", e);
+        }
+    }
+
+    @Override
+    public void shutdown(ScheduledExecutorService threadPool) {
```
rjenkinsjr: Why would we pass a thread pool to this method to shut the pool down, when `SerialClient` does not use a thread pool?

src/main/java/io/auklet/agent/Auklet.java
```diff
@@ -1,109 +1,111 @@
...
+                        client.shutdown(mqttThreadPool);
+                    } catch (Exception e) {
+                        logger.error("Error while shutting down Auklet agent", e);
+                    }
+                })
+        );
     }
 
     public static String getBaseUrl() {
       String fromEnv = System.getenv("AUKLET_BASEURL");
```
npalaska: @shogun656 we might wanna change getBaseUrl to private/protected

src/main/java/io/auklet/agent/Device.java
```diff
@@ -140,7 +140,7 @@ public static String getOrganization(){
 
     public static boolean get_Certs() {
         try {
```
npalaska: Any specific reason why `get_Certs` is public?

src/main/java/io/auklet/agent/Auklet.java
```diff
@@ -1,109 +1,111 @@
...
-            mqttThreadPool.shutdown();
-            try {
-                mqttThreadPool.awaitTermination(3, TimeUnit.SECONDS);
-            } catch (InterruptedException e2) { }
+        logger.info("Directory to store creds: " + folderPath);
+    }
+
+    private static Client createClient(String serialOut, ScheduledExecutorService mqttThreadPool) {
+        if (serialOut.equals("")) {
+            if(Device.get_Certs() && Device.register_device()) {
```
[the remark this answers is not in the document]
shogun656: oh good catch, the constant conflicts hid this from me

src/main/java/io/auklet/agent/broker/SerialClient.java
shogun656: Good point, I will take out the shutdown of the thread pool from the clients and add it as part of the Auklet shutdown so that way we can have the clients more consistent

src/main/java/io/auklet/agent/Auklet.java
shogun656: It needs to be public since I moved the clients to it's package

src/main/java/io/auklet/agent/Device.java
shogun656: no I will change this to protected

PR #38, opened 2018-12-03
Fix constant config refresh

shogun656:


shogun656:
@bleib1dj Yes In the Python Agent, we do update the limits files every hour (this should probably be changed to once a day) and we check if a month has passed to zero out the dataSent

```        
if self.total_samples % self.hour == 0:
        self.emission_rate = self.client.update_limits()
        self.client.check_date()
```

PR #39, opened 2018-12-03
support for an arbitrary dir for config files

npalaska:
- Support for adding arbitrary dir for config files for server type applications. 
- It can be set at runtime using `java -Dauklet.config.dir="custom/file/path/dir" MyApp`
- Additional: 
    - Revert `clientPassword` and `clientId` names to `client_password` and `client_id` respectively since the jsonObject returned by the backend contrains the fields in the later format

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,12 +42,16 @@ public class Demo {
 // where you think the program might exit or inside your own shutdown hook.
 ```
 
-* The agent needs a dedicated folder to create and stores the configuration files on the local machine. 
-Upon every restart the agent tries to create configuration files if they don't already exist. 
-* If the user wants Auklet agent to use specific folder then it can be done using the following command: 
-`java -Dauklet.config.dir="custom/file/path/dir" MyApp`
-* If the `auklet.config.dir` sysProperty is not defined then the agent chooses the directory in the following order
-of its accessibility: working directory -> home directory -> temp directory
+* The agent needs a dedicated folder to create and store its configuration files. The agent will create these files 
+on startup if they do not exist, or will use files that are already available on disk.
+* By default, the agent will use the following JVM system properties, in order, to determine the location from which 
+to read/write its configuration files: `user.dir`, `user.home`, `java.io.tmpdir`. 
+The agent will use the first one of these locations that is writable.
+* If you need to use a specific location for the agent's configuration files other than what is listed above, you can 
+set the JVM system property `auklet.config.dir` or use env variable `AUKLET_CONFIG_DIR` and the agent will attempt to 
+use that value first, assuming it is writable. 
+If it is set to a non-writable directory, or if the property is not set, the agent will fallback to the 
+other properties described above.
 
 # Authorization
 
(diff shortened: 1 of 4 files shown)


src/main/java/io/auklet/agent/Auklet.java
```diff
@@ -59,8 +59,11 @@ public static void exception(Throwable thrown) {
     }
 
     private static void createFolderPath() {
-        folderPath = Util.createCustomFolder("user.dir");
-        if (folderPath == null){
+        folderPath = Util.createCustomFolder("auklet.config.dir");
```
rjenkinsjr: `Util.createCustomFolder` uses `System.getProperty`, which returns `null` if the given property is not defined. This will cause the string concatenation to fail with a `NullPointerException`. If `System.getProperty` returns `null`, then `Util.createCustomFolder` should also return `null`.

README.md
```diff
@@ -41,6 +41,14 @@ public class Demo {
...
 // where you think the program might exit or inside your own shutdown hook.
 ```
+
+* The agent needs a dedicated folder to create and stores the configuration files on the local machine. 
+Upon every restart the agent tries to create configuration files if they don't already exist. 
+* If the user wants Auklet agent to use specific folder then it can be done using the following command: 
+`java -Dauklet.config.dir="custom/file/path/dir" MyApp`
+* If the `auklet.config.dir` sysProperty is not defined then the agent chooses the directory in the following order
+of its accessibility: working directory -> home directory -> temp directory
+
```
rjenkinsjr: Let's use this verbiage:

```
* The agent needs a dedicated folder to create and store its configuration files. The agent will create these files on startup if they do not exist, or will use files that are already available on disk.
* By default, the agent will use the following JVM system properties, in order, to determine the location from which to read/write its configuration files: `user.dir`, `user.home`, `java.io.tmpdir`. The agent will use the first one of these locations that is writable.
* If you need to use a specific location for the agent's configuration files other than what is listed above, you can set the JVM system property `auklet.config.dir` and the agent will attempt to use that value first, assuming it is writable. If it is set to a non-writable directory, or if the property is not set, the agent will fallback to the other properties described above.
```

src/main/java/io/auklet/agent/Util.java
```diff
@@ -67,6 +67,9 @@ protected static String getIpAddress() {
     }
 
     protected static String createCustomFolder(String sysProperty) {
+        if (sysProperty == null || System.getProperty(sysProperty) == null) {
```
rjenkinsjr: I would expect this logic to return `null` if the env var `AUKLET_CONFIG_DIR` is set. The first conditional branch will be false, which will cause `System.getProperty` to be invoked with the _value_ of the env var `AUKLET_CONFIG_DIR`, which will almost certainly not be a proper JVM property and thus will return `null`.

npalaska: Ahh that's right

PR #40, opened 2018-12-04
APM-1697: Fix Mqtt publish error and agent hung problem

npalaska:
- Update the amount of data sent immediately after the client publishes it to mqtt
- This PR makes the `checkDate` timer thread daemon so the agent does not hang after the application exits

### DIFF
--- a/src/main/java/io/auklet/agent/broker/MQTTClient.java
+++ b/src/main/java/io/auklet/agent/broker/MQTTClient.java
@@ -81,11 +81,13 @@ public class MQTTClient implements Client {
 
             @Override
             public void messageArrived(String topic, MqttMessage message) throws Exception {
-                // TODO handle messages received from the Auklet backend.
+                // TODO: handle messages received from the Auklet backend.
             }
 
             @Override
-            public void deliveryComplete(IMqttDeliveryToken token) { }
+            public void deliveryComplete(IMqttDeliveryToken token) {
+                // TODO: handle what happens when MQTT message delivery completes
+            }
         };
     }
 
--- a/src/main/java/io/auklet/agent/broker/MQTTClient.java
+++ b/src/main/java/io/auklet/agent/broker/MQTTClient.java
@@ -103,7 +103,7 @@ public class MQTTClient implements Client {
 
         options.setConnectionTimeout(60);
         options.setKeepAliveInterval(60);
-        options.setCleanSession(true);
+        options.setCleanSession(false);
         options.setAutomaticReconnect(true);
 
         options.setMqttVersion(MqttConnectOptions.MQTT_VERSION_3_1);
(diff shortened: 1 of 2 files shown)


src/main/java/io/auklet/agent/broker/MQTTClient.java
```diff
@@ -186,6 +181,7 @@ public void sendEvent(String topic, byte[] bytesToSend) {
                 message.setQos(1); // At Least Once Semantics
                 client.publish("java/events/" + Device.getOrganization() + "/" +
                         Device.getClientUsername(), message);
+                DataRetention.updateDataSent(message.getPayload().length);
```
rjenkinsjr: If I'm reading this correctly, this new implementation updates the amount of data sent as soon as we _attempt to_ publish data to MQTT, whereas the old implementation would update the amount of data sent as soon as we _receive acknowledgement of_ published data, per QoS 1. I would think we would want the old implementation so as to not penalize end users whose agents fail to publish messages that receive QoS 1 acknowledgements.

Are we changing this implementation for some business use case reason, or did the old implementation simply not work correctly? Or am I reading this wrong entirely and there's something else going on here? @npalaska @bleib1dj

shogun656: @rjenkinsjr In the old implementation, `token.getMessage` would return null which then throws a NullPointerException and then the message is never published. It seems like there is no way to get the size of the message being sent so we have to update the amount the data sent here

rjenkinsjr: @shogun656 the behavior you're describing is 100% in line with what the javadocs say: https://www.eclipse.org/paho/files/javadoc/org/eclipse/paho/client/mqttv3/IMqttDeliveryToken.html#getMessage--.

What I think needs to happen is we need to capture the delivery token ID when we publish (`client.publish` returns an `IMqttDeliveryToken` whose `getMessageId` should always be non-`null`) and associate that in a `Map` to the message size. That way, when the callback's `deliveryComplete` method is invoked, we can correlate the message ID of the token, retrieve the message size from the map and _then_ invoke `DataRetention.updateDataSent`.

rjenkinsjr: @shogun656 just discussed this in the office with @npalaska and @bleib1dj. Updating data usage either on initial publish attempt or upon ACK from the broker is approx. equivalent in terms of the impact on the total data consumption. What we really want is something in between - where we track every _attempt_ to publish and deduct from the data consumption quota for each attempt. The reason for this is so that Auklet's data consumption quota is, as much as possible, a "hard quota" that end users can rely upon for purposes of controlling cellular data consumption/costs.

Therefore, we're going to stick with the implementation in this PR for now and I will shortly make a new ticket to explore tapping into the MQTT library's API further to see if we can get a callback for each attempt at publishing a message, to make our data consumption tracking more accurate.

PR #43, opened 2018-12-13
APM-1695: Refactor codebase

rjenkinsjr:
# Changes

- Converts the majority of the codebase from static methods to objects. `Auklet` remains the entry point with static methods only, and now takes a `Config` object that allows for end users to specify config values in Java code, falling back to env vars and JVM sys props where appropriate.
- Introduces the concept of a *sink* (used to be called a _client_, which was confusing). The sink encapsulates the MessagePack data conversion and JVM system metrics gathering (which comes from the new `OSMX` adapter class).
- Introduces the concept of a *config file*, which previously did not exist. These classes manage the reading of files from disk and fetching from the Auklet API, where required.
- Renamed `io.auklet.agent` package to `io.auklet` for simplicity.
- Added Javadocs.
- Dependency changes:
    - Swapped Apache HttpClient with OkHttp (reduces total number of dependencies by 1 and remove Apache Commons Logging)
    - Swapped `org.json` with `mjson`, because the former's license is not GPL-compatible.
    - Stopped using msgpack's Jackson adapter and now uses `msgpack-core` directly.
    - Made `purejavacomm` an optional (`compileOnly`) dependency, since most users will not need it.
        - Added `sysout-over-slf4j` as an optional dependency, because that's what `purejavacomm` uses for logging
- Added SpotBugs Annotations for compile-time null check support in Code Climate.
- Added Java 9 module export (backwards compatible with Java 8).
    - Upgraded to Gradle Wrapper 4.10.3
- Sealed JAR to prevent malicious attackers from creating classes in our package hierarchies.

## TODO
- [x] Add more debug logging statements where appropriate
- [x] Add package-level Javadocs
- [x] Re-review exception hierarchy/handling
- [x] Consider adding JSR-305 for (non)null annotations
- [x] Add JAR sealing for security
- [x] Determine logic for sink write failures
- [x] Add task queue for `DataUsageTracker.saveUsage`
- [x] Add task queue for `Auklet.send`


src/main/java/io/auklet/Auklet.java
```diff
@@ -0,0 +1,702 @@
...
+    private final boolean uncaughtExceptionHandler;
+    private final String serialPort;
+    private final String macHash;
+    private final String ipAddress;
+    private final OkHttpClient http;
+    private DeviceAuth deviceAuth;
+    private Sink sink;
+    private DataUsageLimit usageLimit;
+    private DataUsageMonitor usageMonitor;
+    private Thread shutdownHook;
```
shogun656: I understand that you want to have these private global variables on top of their private functions but usually it's common standard to have all global variables on top of the class

src/main/java/io/auklet/Auklet.java
```diff
@@ -0,0 +1,694 @@
...
+         * @return {@code this}.
+         */
+        public Config setSerialPort(String serialPort) {
+            this.serialPort = serialPort;
+            return this;
+        }
+
+    }
+
+}
```
shogun656: This class is much too big, I suggest you refactor it into 2 classes, one with all the Auklet stuff in the beginning, one with the Http stuff, and one with the config stuff

src/main/java/io/auklet/Auklet.java
```diff
@@ -0,0 +1,702 @@
...
+            }
+        }
+    }
+
+    /**
+     * <p>Returns the version of the Auklet agent.</p>
+     *
+     * @return never {@code null}. May be the string literal {@code unknown}.
+     */
+    public static String getVersion() { return Auklet.agentVersion; }
```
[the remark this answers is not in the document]
rjenkinsjr: This is purely a matter of style. We don't have a standard yet for this repo and I'm not sure yet which direction I want to take.

src/main/java/io/auklet/config/DataUsageTracker.java
```diff
@@ -0,0 +1,110 @@
...
+        new Thread(() -> {
+            synchronized (this.lock) {
+                try {
+                    this.writeToDisk(this.bytesSent);
+                } catch (IOException | SecurityException e) {
+                    LOGGER.warn("Could not save data usage to disk", e);
+                }
+            }
+        }).start();
+    }
```
shogun656: Why asynchronously save the data to a file instead of synchronously. You are locking when you change `bytesSent` and create this thread but once you create this thread you get out of the lock. Don't you want to wait until the `bytesSent` is saved to a file before getting out of the lock?

src/main/java/io/auklet/Auklet.java
```diff
@@ -0,0 +1,702 @@
...
+                Runtime.getRuntime().addShutdownHook(shutdownHook);
+            }
+            // Enable the uncaught exception handler, if requested.
+            if (agent.uncaughtExceptionHandler) {
+                Thread.setDefaultUncaughtExceptionHandler(new AukletExceptionHandler());
+            }
+            // Set the global agent object.
+            Auklet.agent = agent;
+        }
+    }
```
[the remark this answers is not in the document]
rjenkinsjr: Some of this is leftover from when I was in the middle of the refactoring and needed to keep track of all the various functionalities I was porting. I'll go through and clean up unnecessary comments.

src/main/java/io/auklet/Auklet.java
rjenkinsjr: This is also a matter of style, but you are absolutely correct that normally variables go at the beginning of the class declaration.

The reason I didn't do this is because this class is designed as a singleton with static entry points, so I wanted to separate the static members from the non-static members, as I felt it made the class easier to comprehend.

I will address your other comment about class size first and then we can revisit this.

src/main/java/io/auklet/Auklet.java
rjenkinsjr: I think you just suggested a total of 3 classes, not 2. :)

Before I opened this PR I considered making `Auklet.Config` a separate top-level class. I chose not to because doing it the way I did it makes the private variables accessible in the `Auklet` class, which eliminates the need for getter methods (in reality, `Auklet.Config` is really a write-only API for the end user). In retrospect, this probably is an unnecessary design choice, so I'll move that class out.

With regard to HTTP - I want to have a single HTTP client that is used for all communication with the API; the current agent instantiates new client objects for each request, which is inefficient. Since `Auklet` is the only singleton in the codebase (except `OSMX` but that's not relevant here), it made sense to me to put it here, but in retrospect I think that functionality should be extracted as you suggest, so I'll move that to another class too.

src/main/java/io/auklet/misc/AukletApi.java
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * <p>Makes an authenticated request to the Auklet API.</p>
+     *
+     * @param request never {@code null}.
+     * @return never {@code null}.
+     * @throws AukletException if an error occurs with the request.
+     */
+    @NonNull public Response doRequest(@NonNull Request.Builder request) throws AukletException {
+        if (request == null) throw new AukletException("HTTP request is null");
```
shogun656: Isn't this redundant when you have `@NonNull` in the argument of the function.

src/main/java/io/auklet/jvm/OSMX.java
```diff
@@ -0,0 +1,149 @@
...
+ * MXBean returns a value that is considered exceptional (e.g. when
+ * {@link java.lang.management.OperatingSystemMXBean#getSystemLoadAverage()} returns a negative value). Check the
+ * Javadocs for the above classes for details on exceptional return values. Methods that would otherwise throw a
+ * {@link SecurityException} will instead return {@link Optional#empty()}.</p>
+ *
+ * <p>This is a singleton; use {@link #BEAN} to retrieve the instance.</p>
+ */
+public enum OSMX {
+
+    BEAN;
```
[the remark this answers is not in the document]
rjenkinsjr: Enums in Java are compile-time-enforced singleton classes with a fixed number of instances of the class. It's like Scala's `object` syntax except you can have more than once instance.

This syntax is what instantiates the class at compile time. It is a syntax error not to have this syntax in the definition of an enum class, because otherwise you would have an enum with no instances, which is pointless.

You can have as many instances of an enum as you want. Only one instance is needed for this class, since it is serving as a frontend to the JVM's OperatingSystemMXBean, and only one such bean is guaranteed to exist in any running JVM (per Javadocs). I chose the name `BEAN` so as to produce the fluent syntax `OSMX.BEAN.methodName()`.

See https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html for more info on enums. If I was a betting man, without looking at source code, I'd say Scala `object`s are built on Java enums.

shogun656:
If the javadoc tool will return warnings about not documenting all the parameters than it may be better to leave them documented even though it is redundant. It's best to stick with best practices even though it can be a little redundant

src/main/java/io/auklet/config/AbstractConfigFile.java
```diff
@@ -0,0 +1,31 @@
+package io.auklet.config;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import io.auklet.Auklet;
```
npalaska: @rjenkinsjr based on the documentation http://findbugs.sourceforge.net/api/edu/umd/cs/findbugs/annotations/package-summary.html#package_description, annotations for FindBugs are mostly deprecated, shouldn't we be using javax.annotation instead?

src/main/java/io/auklet/jvm/OSMX.java
```diff
@@ -0,0 +1,149 @@
...
+    // deserialization taking place.
+    static {
+        realBean = ManagementFactory.getOperatingSystemMXBean();
+        boolean sun = false;
+        boolean sunUnix = false;
+        try {
+            sun = realBean instanceof com.sun.management.OperatingSystemMXBean; // NOSONAR
+        } catch (NoClassDefFoundError e) {
+            LOGGER.warn("com.sun.management.OperatingSystemMXBean does not exist; system memory and JVM CPU usage stats will be available.");
+        }
```
npalaska: @rjenkinsjr slightly confused here, if package com.sun.management.OperatingSystemMXBean does not exist, system memory and JVM CPU usage stats will not be available isn't it?

rjenkinsjr: Good catch, this is a typo. :)

src/main/java/io/auklet/config/AbstractConfigFile.java
rjenkinsjr: As I mentioned to @shogun656 in his review, the reason I added these annotations was because I wanted null check detection during our static code analysis via SonarJava in Code Climate. The two most prevalent null annotation implementations that SonarJava recognizes are FindBugs/SpotBugs (`edu.umd.cs.findbugs.annotations`) and the `javax` annotations from the JSR-305 initiative/reference implementation that was once stored in Google Code.

Your point is valid that the FindBugs/SpotBugs annotations are deprecated, but they were deprecated in anticipation of JSR-305 becoming an approved/adopted Java standard. There are two problems with JSR-305:
- The JSR approval never happened, and it is currently defunct with no further development expected
- The annotations defined in that JSR are in the `javax` package but are not developed by Oracle. Depending on how you read the Oracle Java Binary License, it may not be legal to distribute an app containing the JSR-305 annotations alongside an Oracle JRE/JDK. This limitation would not affect us directly, but could affect our customers depending on how they choose to write/distribute their code.
- The JSR-305 annotations have a retention policy of `RUNTIME`, which means that the bytecode that defines the annotation (the `.class` file) must be on the classpath or any classes that reference that annotation cannot be loaded by the JVM, which would cause the Auklet agent to emit unnecessary compile-time warnings and/or (more severe) break at runtime.

In contrast to the JSR-305 annotations, the FindBugs/SpotBugs annotations have the retention policy `CLASS`, which means it is syntactically legal for the annotations to be missing at runtime, despite the ironic fact that the FindBugs/SpotBugs annotations depend upon the JSR-305 annotations. This gives us the annotations/behavior we want without causing any potentially murky legal issues for our customers down the road.

**tl;dr**: FindBugs annotations were deprecated prematurely, and the only other alternative recognized by SonarJava poses possible legal distribution challenges for our customers.

rjenkinsjr: @npalaska see here for details related to my answer above: https://stackoverflow.com/a/36198568

npalaska: @rjenkinsjr thanks for the clear explanation and a very strong 2nd point we do not want the customers to get in some legal issues

src/main/java/io/auklet/config/AbstractConfigFile.java
```diff
@@ -0,0 +1,32 @@
...
+ * configuration directory.</p>
+ */
+@NotThreadSafe
+public abstract class AbstractConfigFile extends HasAgent {
+
+    protected File file;
+
+    @Override public void start(@NonNull Auklet agent) throws AukletException {
+        super.start(agent);
+        this.file = new File(agent.getConfigDir(), this.getName());
```
shogun656: Do you think we should check if the file exist and if not create a new file here?

src/main/java/io/auklet/config/AukletIoBrokers.java
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * <p>Returns the MQTT connection URL</p>
+     *
+     * @return never {@code null}.
+     */
+    @NonNull public String getUrl() { return this.url; }
+
+    @Override protected Json readFromDisk() {
+        try {
+            return this.validate(Json.read(this.getStringFromDisk()));
```
shogun656: Before doing this, you should check if `this.getStringFromDisk()` returns null as this could be the first time you are writing to this file. I was running into that problem when working with the java agent. I ended up adding this to all `readFromDisk` functions on all classes in the config package
```
            String st = this.getStringFromDisk();
            if(st == null || st.isEmpty()) return null;
            return Json.read(st);```

File: src/main/java/io/auklet/config/AbstractConfigFileFromApi.java [hunk 12]
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * <p>Loads the config file from disk into a string, using the UTF-8 charset.</p>
+     *
+     * @return never {@code null}.
+     * @throws IOException if the file does not exit or cannot be read.
+     */
+    @NonNull protected final String getStringFromDisk() throws IOException {
+        try {
+            byte[] bytes = Util.read(this.file);
+            return new String(bytes, "UTF-8");
```
shogun656: We should check if the file is empty and return null if it is empty here. Something like `if(this.file.length() == 0) return null;`

src/main/java/io/auklet/sink/AbstractSink.java
```diff
@@ -0,0 +1,144 @@
...
+                        .packString("excType").packString(throwable.toString())
+                        .packString("stackTrace").packMapHeader(stackTrace.length);
+                for (StackTraceElement ste : stackTrace) {
+                    int lineNumber = ste.getLineNumber();
+                    this.msgpack
+                            .packString("functionName").packString(ste.getMethodName())
+                            .packString("className").packString(ste.getClassName())
+                            .packString("filePath").packString(Util.defaultValue(ste.getFileName(), ""))
+                            // Normalize all negative return values.
+                            .packString("lineNumber").packInt(lineNumber < 0 ? -1 : lineNumber);
```
shogun656: We want to stackTrace to be an array of maps so we should do 
```
                this.msgpack
                        .packString("timestamp").packLong(System.currentTimeMillis())
                        .packString("excType").packString(name)
                        .packString("stackTrace").packArrayHeader(stackTrace.length);
                for (StackTraceElement ste : stackTrace) {
                    int lineNumber = ste.getLineNumber();
                    this.msgpack.packMapHeader(4)
                            .packString("functionName").packString(ste.getMethodName())
                            .packString("className").packString(ste.getClassName())
                            .packString("filePath").packString(Util.defaultValue(ste.getFileName(), "unknown"))
                            // Normalize all negative return values.
                            .packString("lineNumber").packInt(lineNumber < 0 ? -1 : lineNumber);
                }```

File: src/main/java/io/auklet/config/AbstractConfigFile.java (hunk 10 above)
Comment (rjenkinsjr) [reply src/main/java/io/auklet/config/AbstractConfigFile.java]: Every concrete implementation of this class uses either `Util.writeUtf8` or `Util.write` to write the file contents to disk. Both of these methods overwrite whatever file may already exist.

Because of this, I don't see any reason why we would want blank files. In fact, blank files would cause runtime errors with any JSON config files because they would fail schema validation.

If you have a use case for having the files on disk but empty, I'm open to suggestions, but as it stands right now I'm not sure what value add there is in doing so.

File: src/main/java/io/auklet/config/AukletIoBrokers.java (hunk 11 above)
Comment (rjenkinsjr) [reply src/main/java/io/auklet/config/AukletIoBrokers.java]: `getStringFromDisk()` is defined to return `@NonNull`. If it were to return `null` that would be a programming error, not a valid condition that should be checked.

You said "this could be the first time you are writing to this file":
1. This logic is handling file reads, not writes.
1. Why would/should we expect `null` in the first place?

File: src/main/java/io/auklet/sink/AbstractSink.java (hunk 13 above)
Comment (rjenkinsjr) [reply src/main/java/io/auklet/sink/AbstractSink.java]: Nice find, that would have been a big deal. :)

File: src/main/java/io/auklet/config/AukletIoBrokers.java (hunk 11 above)
Comment (rjenkinsjr) [reply src/main/java/io/auklet/config/AukletIoBrokers.java]: In the current implementation in this PR, `getStringFromDisk()` calls `Util.read(this.file)`, which returns an empty array if the file does not exist.

I just pushed some changes, including a short-circuit check inside `Util.read` on `file.length()`, which according to Javadocs returns `0` if the file does not exist. If the behavior you're reporting existed in this PR prior to my most recent commits, it should be fixed now, unless Android has some oddball implementation of `java.io.File` that doesn't respect the Javadocs.

**Comment by shogun656** [comment]
I'm going to try this PR locally to see if I can find any bugs real quick

File: src/main/java/io/auklet/config/AbstractConfigFile.java (hunk 10 above)
Comment (npalaska) [reply src/main/java/io/auklet/config/AbstractConfigFile.java]: @rjenkinsjr, am not very familiar with `mjson` package but can it automatically understand the json schema and not have us predefine the schema for every config

File: src/main/java/io/auklet/Auklet.java [hunk 14]
```diff
@@ -0,0 +1,558 @@
...
+                System.getProperty("auklet.config.dir"),
+                System.getProperty("user.dir"),
+                System.getProperty("user.home"),
+                System.getProperty("java.io.tmpdir")
+        );
+        // Drop any env vars/sysprops whose value is null, and append the auklet subdir to each remaining value.
+        List<String> filteredConfigDirs = new ArrayList<>();
+        for (String dir : possibleConfigDirs) {
+            if (!Util.isNullOrEmpty(dir)) filteredConfigDirs.add(Util.removeTrailingSlash(dir) + "/aukletFiles");
+        }
```
npalaska: @rjenkinsjr ~~I think we need to include a `break` statement in the for loop isnt it?~~

src/main/java/io/auklet/Auklet.java
```diff
@@ -0,0 +1,558 @@
...
+        // If a directory contains the auth file, use that directory.
+        // We don't care if the other files don't exist because we'll create them later if needed.
+        LOGGER.debug("Checking directories for existing config files.");
+        for (String dir : filteredConfigDirs) {
+            File authFile = new File(dir, DeviceAuth.FILENAME);
+            try {
+                if (authFile.exists()) {
```
npalaska: @rjenkinsjr we check if the directory contains the authFile, however, what if the `temp` dir contains the authFile do we want to continue using the `temp` dir without checking if any other fileSystem is accessible this time? Do you think we also need to check if the dir contains cert files?

src/main/java/io/auklet/Auklet.java
```diff
@@ -0,0 +1,558 @@
...
+                    if (agent == null) {
+                        LOGGER.debug("Ignoring send request because agent is null.");
+                        return;
+                    }
+                    agent.doSend(throwable);
+                }
+            }
+        };
+        try {
+            DAEMON.submit(sendTask);
```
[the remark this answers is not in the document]
rjenkinsjr: `initTask` and `sendTask` are themselves separate tasks. It is true that they are executed within the same thread, but that fact is merely an implementation detail and has no bearing on the relationship between the two tasks.

The only relationship between them is that they share a lock object, to prevent attempting to send data before the agent is initialized. One task does not execute "within" another task.

src/main/java/io/auklet/Auklet.java
rjenkinsjr: After reviewing this comment and the relevant logic, we need to discuss this further. I'll reach out to you on Slack to figure out what to do here.

src/main/java/io/auklet/core/AukletApi.java
```diff
@@ -0,0 +1,114 @@
...
+                this.httpClient.connectionPool().evictAll();
+                this.httpClient.cache().close();
+            } catch (IOException e) {
+                LOGGER.warn("Error while shutting down Auklet API.", e);
+            }
+        }
+    }
+
+    /**
+     * <p>Creates an OkHttp logging interceptor that sends all HTTP logs to the SLF4J logger named
```
bleib1dj: Do you want the `Http...` on a new line?

src/main/java/io/auklet/core/HasAgent.java
```diff
@@ -0,0 +1,59 @@
...
+    private Auklet agent = null;
+
+    /**
+     * <p>Sets the Auklet agent reference and invokes the post-construction start logic for this object.</p>
+     *
+     * @param agent the Auklet agent object. Never {@code null}.
+     * @throws AukletException if the agent reference has already been set, or if the input is {@code null}.
+     */
+    public void start(@NonNull Auklet agent) throws AukletException {
+        if (agent == null) throw new AukletException("Auklet agent is null.");
```
npalaska: Not super imp but just a thought - I feel instead of `start`, `check` would have been a better choice since here `start` does not start anything per se but just checks the auklet agent against null

rjenkinsjr: We need a `start` method to provide `post-construction start logic`. Perhaps putting the agent reference setter in here is overloading the true purpose of the method. I'll look into refactoring this to make it more clear.

PR #48, opened 2019-01-22
APM-1724: Add support for Android

shogun656:
This PR adds support for Android. There is going to be some permissions the end user will need to give and the end user will need to give their application context.


src/main/java/io/auklet/core/AndroidMetrics.java
```diff
@@ -0,0 +1,97 @@
...
+        // On Android 8+, Google has restricted access to the proc files
+        if (Build.VERSION.SDK_INT < 26) {
+            cpuThread =  new ScheduledThreadPoolExecutor(1, Util.createDaemonThreadFactory("Android-Metrics"));
+            cpuThread.scheduleAtFixedRate(this.calculateCPUUsage(), 0L, 1L, TimeUnit.SECONDS);
+        } else {
+            LOGGER.debug("We are unable to attain CPU data on Devices running Android 8+");
+        }
+    }
+
+    private Runnable calculateCPUUsage() {
```
rjenkinsjr: What's the rationale for using a `Runnable`/`Thread` for obtaining the CPU info? For Java apps, we obtain performance data only when handling an exception, rather than on a periodic basis. Do we need to handle this differently for Android?

build.gradle
```diff
@@ -38,6 +39,7 @@ dependencies {
     compileOnly "com.github.spotbugs:spotbugs-annotations:3.1.10"
     compileOnly "com.github.purejavacomm:purejavacomm:1.0.1.RELEASE"
     compileOnly "uk.org.lidalia:sysout-over-slf4j:1.0.2"
+    compileOnly "com.google.android:android:4.1.1.4"
```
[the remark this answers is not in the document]
shogun656: Yes, this is equivalent to API level 16, the lowest version we are able to support

src/main/java/io/auklet/core/AndroidMetrics.java
shogun656: So this is the only way to receive cpu percentage for Linux based systems. We need previous data to do our calculations for cpu percentage since we are unable to have previous data if we wait until an error occurred. I'm pretty sure the library we use for Java spins up it's own thread to do this as well

src/main/java/io/auklet/platform/Platform.java
```diff
@@ -0,0 +1,35 @@
...
+     * @throws IllegalArgumentException if the MessagePacker is {@code null}.
+     */
+    void addSystemMetrics(@NonNull MessagePacker msgpack) throws IOException;
+
+    /**
+     * <p>Returns whether the current system is Android.</p>
+     *
+     * @return never {@code null}.
+     */
+    @NonNull Boolean isAndroid();
```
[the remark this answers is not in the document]
shogun656: No, I left in there in case we needed it, but I doubt we will need it since we have the platform architecture so I will remove it

src/main/java/io/auklet/core/AndroidMetrics.java
```diff
@@ -36,11 +35,14 @@
     private Long workDiff;
     private float cpuUsage = 0;
 
-    public void start(@NonNull Auklet agent) throws AukletException {
-        this.setAgent(agent);
-        manager = (ActivityManager) agent.getContext().getSystemService(Context.ACTIVITY_SERVICE);
+    public AndroidMetrics(Context context) {
```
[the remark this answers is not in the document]
shogun656: I think `AndroidMetrics` merits its own class since it is specific in getting the metrics for android. Throwing that functionality into the `AndroidPlatform` class seems like it would clutter the `AndroidPlatform` class

src/main/java/io/auklet/Auklet.java
```diff
@@ -518,6 +498,7 @@ private void start() throws AukletException {
         this.deviceAuth.start(this);
         this.usageMonitor.start(this);
         this.sink.start(this);
+        this.platform.start(this);
```
rjenkinsjr: I know it doesn't make any functional difference as of right now, but conceptually it seems that the platform should be started earlier. I'd like to establish a sensible hierarchy so that if there are any interdependencies in the future we won't have to think so much about it. Seems to me that it would be best for the platform to start after the usage monitor but before the sink. What do you think?

src/main/java/io/auklet/core/AndroidMetrics.java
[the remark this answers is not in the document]
rjenkinsjr: In that case, I think we need to do some package juggling so we don't overload `io.auklet.core` and to better differentiate the purposes of these classes. I propose the following, let me know what you think:
- Move `OSMX` and `AndroidMetrics` into a new package `io.auklet.platform.metrics`.
- Move `AukletExceptionHandler` into `io.auklet.core` and delete the package `io.auklet.jvm`.
- Move `Util` and `PahoLogger` into a new package `io.auklet.misc`.

Also, since I meant to make this original comment on the class and not the constructor, I just noticed that this constructor is missing a Javadoc.

src/main/java/io/auklet/core/AndroidMetrics.java
```diff
@@ -0,0 +1,93 @@
...
+ * <p>CPU Usage can only be retrieved for devices running on Android 7 or older. If your device is running on
+ * something newer than Android 7 than the CPU Usage will return 0.</p>
+ */
+public final class AndroidMetrics extends HasAgent {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AndroidMetrics.class);
+    private ActivityManager.MemoryInfo memInfo = new ActivityManager.MemoryInfo();
+    private ActivityManager manager;
+
+    private Long total;
```
rjenkinsjr: Should these 6 `Long`s be objects or primitives? Is it possible/sensible for any of them to be `null`?

src/main/java/io/auklet/core/AndroidMetrics.java
```diff
@@ -0,0 +1,93 @@
...
+    private Long workDiff;
+    private float cpuUsage = 0;
+
+    public AndroidMetrics(Context context) {
+        manager = (ActivityManager) context.getSystemService(Context.ACTIVITY_SERVICE);
+        manager.getMemoryInfo(memInfo);
+        total = totalBefore = totalDiff = work = workBefore = workDiff = 0L;
+    }
+
+    public void start(@NonNull Auklet agent) throws AukletException {
```
rjenkinsjr: Since this class is now an implementation detail of `AndroidPlatform`, it seems like overkill to me to have `start` logic for this class, or to have a reference to the Auklet agent. What if we did this instead?
- Do the `SDK_INT` check in the constructor, and log the message if appropriate.
- Have `calculateCPUUsage()` possibly return `null` and be invoked by `AndroidPlatform.start()`, so that the latter can decide whether or not to schedule the task?

src/main/java/io/auklet/core/AndroidMetrics.java
```diff
@@ -0,0 +1,93 @@
...
+
+        // On Android 8+, Google has restricted access to the proc files
+        if (Build.VERSION.SDK_INT < 26) {
+            agent.scheduleRepeatingTask(this.calculateCPUUsage(), 0L, 1L, TimeUnit.SECONDS);
+        } else {
+            LOGGER.debug("We are unable to attain CPU data on Devices running Android 8+");
+        }
+    }
+
+    private Runnable calculateCPUUsage() {
```
rjenkinsjr: This method, as well as `getCPUUsage`, should be capitalized as `*CpuUsage` for consistent usage of camel-case.

src/main/java/io/auklet/platform/AndroidPlatform.java
```diff
@@ -0,0 +1,38 @@
...
+    private final Context context;
+    private final AndroidMetrics metrics;
+
+    public AndroidPlatform(Context context) throws AukletException {
+        this.context = context;
+        this.metrics = new AndroidMetrics(context);
+    }
+
+    @Override public void start(@NonNull Auklet agent) throws AukletException {
+        this.setAgent(agent);
```
rjenkinsjr: This method does everything that the superclass method does. Either call `super.start(agent)` or remove the method definition in the superclass (in which case you'd have to define it in `JavaPlatform`).

src/main/java/io/auklet/platform/AbstractPlatform.java
```diff
@@ -0,0 +1,93 @@
...
+     * <p>Returns the directory the Auklet agent will use to store its configuration files. This method
+     * creates/tests write access to the target config directory after determining which directory to use,
+     * per the logic described in the class-level Javadoc.</p>
+     *
+     * @param fromConfig the value from the {@link Config} object, env var and/or JVM sysprop, possibly
+     * {@code null}.
+     * @return possibly {@code null}, in which case the Auklet agent must throw an exception during
+     * initialization and all data sent to the agent must be silently discarded.
+     */
+    @CheckForNull public File obtainConfigDir(@Nullable String fromConfig) {
```
rjenkinsjr: This method should be defined in the `Platform` interface. The implementation here should remain here, but be declared `final` so that subclasses cannot override it.

After changing this, you can make the `AbstractPlatform` variable in the `Auklet` class a `Platform` instead.

src/main/java/io/auklet/platform/Platform.java
```diff
@@ -0,0 +1,28 @@
...
+public interface Platform {
+
+    /**
+     * <p>Determines the possible config directories Auklet can use based on the underlying platform.</p>
+     *
+     * @param fromConfig the value from the {@link Config} object, env var and/or JVM sysprop, possibly
+     * {@code null}.
+     * @return the list of possible config directories that we are able to use
+     */
+    @NonNull List<String> getPossibleConfigDirs(@Nullable String fromConfig);
```
rjenkinsjr: This method is an implementation detail of any given `Platform`, so it should now be made a `protected abstract` method in the `AbstractPlatform` class.

src/main/java/io/auklet/platform/Platform.java
```diff
@@ -0,0 +1,28 @@
...
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import io.auklet.Config;
+import org.msgpack.core.MessagePacker;
+
+import java.io.IOException;
+import java.util.List;
+
+public interface Platform {
```
rjenkinsjr: Needs Javadoc.

src/main/java/io/auklet/platform/AndroidPlatform.java
```diff
@@ -0,0 +1,38 @@
...
+        this.metrics.start(agent);
+    }
+
+    @Override public List<String> getPossibleConfigDirs(@Nullable String fromConfig) {
```
rjenkinsjr: Even though it's an `@Override`, this method merits its own Javadoc because it ignores the input argument and always returns a list with exactly 1 element. Please add a Javadoc that describes both this behavior and the fact that the input argument is unused.

src/main/java/io/auklet/platform/AndroidPlatform.java
```diff
@@ -0,0 +1,38 @@
...
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+
+public class AndroidPlatform extends AbstractPlatform {
+
+    private final Context context;
+    private final AndroidMetrics metrics;
+
+    public AndroidPlatform(Context context) throws AukletException {
```
rjenkinsjr: Needs Javadoc.

src/main/java/io/auklet/core/AndroidMetrics.java
```diff
@@ -0,0 +1,93 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * <p>This class handles retrieving the Memory and CPU Usage for Android devices.</p>
```
rjenkinsjr: Throughout your Javadocs (not just here, not sure where else this appears), let's not capitalize `Memory` or `Usage` in `CPU Usage`, as neither of them are proper pronouns.

src/main/java/io/auklet/core/AndroidMetrics.java
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * <p>CPU Usage can only be retrieved for devices running on Android 7 or older. If your device is running on
+ * something newer than Android 7 than the CPU Usage will return 0.</p>
```
rjenkinsjr: Should be written as `... Android 7, then the ...`.

src/main/java/io/auklet/core/AndroidMetrics.java
shogun656: No they can never be null, so does that mean it better to have them as primitives

rjenkinsjr: Yes. Using the object form implies that they could be `null` and in fact may throw false positive during static code analysis. Making them primitives is self-documenting and eliminates the possibility of `null`.

src/main/java/io/auklet/platform/AndroidPlatform.java
shogun656: 'start(agent)' is not defined in the super class `HasAgent`. It's an abstract function there

rjenkinsjr: Correct, but it's defined in `AbstractPlatform`.

src/main/java/io/auklet/core/AndroidMetrics.java
shogun656: I'm not sure if I understand this correctly. I see you want me to do the `SDK_INT` check in the constructor of the AndroidMetrics class but I won't have access to the Auklet class to start the task. Also, why would `calculateCPUUsage` ever return null?

rjenkinsjr: I should have elaborated on my suggestion more completely. Here's what I'm thinking:
- Move the logging statement to the `AndroidMetrics` constructor.
- Make `calculateCPUUsage()` `public` and make it return `null` if the SDK level is too high.
- In `AndroidPlatform.start()`, instead of invoking `this.metrics.start(agent)`, invoke `this.metrics.calculateCPUUsage()`. If it returns non-`null`, pass it to `agent.scheduleRepeatingTask()`.

As a result of these changes, `AndroidMetrics` would no longer have a `start` method, and would no longer `extends HasAgent`.

src/main/java/io/auklet/platform/AbstractPlatform.java
shogun656: I made this change but in Auklet we are still going to need leave the variable in the `Auklet` class as a `AbstractPlatform` since we call `platform.start()` in the `Auklet` class

rjenkinsjr: Good point, forgot about that. That's fine.

src/main/java/io/auklet/Auklet.java
```diff
@@ -92,13 +95,19 @@
      * @throws IllegalStateException if the agent is already initialized.
      * @throws AukletException if the agent cannot be initialized.
      */
-    private Auklet(@Nullable Config config) throws AukletException {
+    private Auklet(@Nullable Context context, @Nullable Config config) throws AukletException {
```
rjenkinsjr: 1. New parameter `context` is not documented.
2. If `context` is `null` and `this.serialPort` is not `null` (line 124), that means someone is trying to use a serial port sink on Android, which makes no sense. Please add a check for this and throw an `AukletException` so that the agent init will fail.

shogun656:
@rjenkinsjr It's tested and working on my end

PR #60, opened 2019-03-22
Hide MQTT disconnect errors

rjenkinsjr:
This is a workaround for https://github.com/aukletio/Auklet-Agent-Java/pull/57.

Changes:
- MQTT disconnect exception is not logged.
- `CancellationException`s are not logged while the agent is shutting down. Sometimes MQTT disconnect causes these exceptions if their corresponding thread tasks do not complete before the JVM begins to shutdown. In general, it's good not to log _any_ `CancellationException`s during agent shutdown (because it makes sense for tasks to be cancelled during that timeframe), so this change will _not_ be reverted when https://github.com/aukletio/Auklet-Agent-Java/pull/57 is merged, and thus this PR is labeled as an enhancement.

### DIFF
--- a/src/main/java/io/auklet/sink/AukletIoSink.java
+++ b/src/main/java/io/auklet/sink/AukletIoSink.java
@@ -20,7 +20,6 @@ import java.io.IOException;
 import java.security.*;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
-import java.util.concurrent.ScheduledExecutorService;
 
 /** <p>The default Auklet data sink, which sends data to {@code auklet.io} via MQTT.</p> */
 @ThreadSafe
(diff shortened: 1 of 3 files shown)


src/main/java/io/auklet/sink/AukletIoSink.java
```diff
@@ -83,11 +82,15 @@
...
                 if (this.client.isConnected()) {
+                    this.executorService.logCancelExceptions(false);
                     try {
                         // Wait 2 seconds for work to quiesce and 1 second for disconnect to finish.
                         this.client.disconnect(2000L).waitForCompletion(1000L);
                     } catch (MqttException e) {
-                        LOGGER.warn("Error while disconnecting MQTT client.", e);
+                        // TODO remove this conditional after upgrading Paho to fix this
+                        if (!e.getMessage().equals("Timed out waiting for a response from the server")) {
+                            LOGGER.warn("Error while disconnecting MQTT client.", e);
```
npalaska: @rjenkinsjr So once the Paho package upgrades, we don't have to hardcode this message based condition right?

rjenkinsjr: Correct, we would remove this conditional once we merge #57 and start logging all exceptions again.

PR #61, opened 2019-03-22
Fix retrieval of agent version

rjenkinsjr:
Not sure when this broke (or if it ever worked), but the agent version has not been retrieved correctly from the JAR manifest's `Implementation-Version` attribute. This approach, even when fixed for Java SE, would not work for Android.

This PR adds a Gradle plugin that generates a `BuildConfig` Java class during the build with our version number, which we then reference at runtime to get the agent version. This fixes both Java SE and Android.


npalaska:
@rjenkinsjr not sure if we have ever done testing specifically for this functionality

PR #63, opened 2019-03-25
Miscellaneous fixes/cleanup

rjenkinsjr:
- Cleanup Javadocs, documentation of thread safety/nullable handling, and code style/formatting.
- Change how some exceptions were being propagated internally.
- Fix thread safety for `DataUsageTracker` and `AndroidMetrics`.
- Cause agent init to fail on Android if API level is less than 16.
- Fix memory usage retrieval in `AndroidMetrics`.


npalaska:
I did the testing on this PR except for `AndroidMetrics ` and seems to be working

PR #72, opened 2019-04-01
Datapoint Integration

tylerwiersing:


src/main/java/io/auklet/Auklet.java
```diff
@@ -262,6 +264,225 @@ public static void send(@Nullable final Throwable throwable) {
...
+     * <p>This method sends a datapoint map (non-array) with a single element whose key is "value"</p>
+     *
+     * @param dataType User specified string which describes the provided data
+     * @param data A {@code boolean} which is the data described by the dataType
+     */
+    public void send(String dataType, boolean data) {
+        LOGGER.debug("Attempting to build datapoint object.");
+        try {
+            Datapoint datapoint = new Datapoint(data);
+            this.send(dataType, datapoint);
```
rjenkinsjr: You can just say `this.send(dataType, new Datapoint(data))` for brevity.

src/main/java/io/auklet/Auklet.java
```diff
@@ -504,6 +725,28 @@ private void doSend(@Nullable final Throwable throwable) {
         }
     }
 
+    /**
+     * <p>Queues a task to submit the given datapoint to the data sink.</p>
+     * @param data
+     * @param dataType
+     */
+    private void doSend(@NonNull final String dataType, @NonNull final io.auklet.Datapoint data) {
```
rjenkinsjr: No need to use the fully qualified class name `io.auklet.Datapoint` since it's in the same package.

src/main/java/io/auklet/Auklet.java
```diff
@@ -504,6 +725,28 @@ private void doSend(@Nullable final Throwable throwable) {
...
+        try {
+            this.scheduleOneShotTask(new Runnable() {
+                @Override public void run() {
+                    try {
+                        LOGGER.debug("Sending datapoint: {}-{}", dataType, data);
```
rjenkinsjr: We don't want to log the data itself, as it might be large and we also don't log exceptions that we send to Auklet either. We should log only the data type and the length of the data here.

src/main/java/io/auklet/Datapoint.java
```diff
@@ -0,0 +1,246 @@
...
+import io.auklet.AukletException;
+import org.msgpack.core.MessageBufferPacker;
+import org.msgpack.core.MessagePack;
+import net.jcip.annotations.GuardedBy;
+
+import java.io.*;
+import java.util.*;
+
+
+public class Datapoint {
```
rjenkinsjr: This class should be `final` and `@Immutable`.

src/main/java/io/auklet/Datapoint.java
```diff
@@ -0,0 +1,246 @@
...
+
+    @GuardedBy("itself") protected final MessageBufferPacker msgpack = MessagePack.newDefaultBufferPacker();
```
rjenkinsjr: This field should not exist because it's only used in the constructor. You can just use a variable in the constructor and continue storing the final result in the `dataValue` array (to be replaced with a `Value` variable).

src/main/java/io/auklet/Datapoint.java
```diff
@@ -0,0 +1,246 @@
...
+    private final String dataType = "value";
```
rjenkinsjr: This variable name confused me at first. It's not a representation of the datapoint's type - it's a string literal we use for when marshalling simple values (non-`List` and non-`Map`) into a MessagePack map. This should be a `static` field named `SIMPLE_KEY`.

src/main/java/io/auklet/Datapoint.java
```diff
@@ -0,0 +1,246 @@
...
+    public byte[] dataValue;
```
rjenkinsjr: This field should be `private` and `final`, and a `getValue()` method should be created that returns a *copy* of this array (merely returning the array opens up the possibility of unintentional modification, even when declared `final`).

It should also be of type `org.msgpack.value.Value` so that we can use `ValueFactory` to properly convert everything, and then use its `writeTo()` method to write the value to `AbstractSink`'s `MessagePacker`.

src/main/java/io/auklet/Datapoint.java
```diff
@@ -0,0 +1,246 @@
...
+
+    /**
+     * <p>Initializes a Datapoint class which packs up the passed data into messagepack byte arrays</p>
+     *
+     * @param data data to be packed
+     * @throws AukletException to wrap any underlying exceptions.
+     */
+    public Datapoint(boolean data) throws AukletException {
+        synchronized (this.msgpack) {
```
rjenkinsjr: There is no need to synchronize inside a constructor when accessing your own class members. Java guarantees that objects under construction are not visible to other objects (unless you leak `this`, which you're not doing). Also, given that `msgpack` doesn't need to be a class member, there'd be nothing to synchronize on.

src/main/java/io/auklet/Datapoint.java
```diff
@@ -0,0 +1,246 @@
...
+     *
+     * @param data data to be packed
+     * @throws AukletException to wrap any underlying exceptions.
+     */
+    public Datapoint(List data) throws AukletException {
+        synchronized (this.msgpack) {
+            this.msgpack.clear();
+            try {
+                this.msgpack.packArrayHeader(1)
+                        .packString(this.dataType).packArrayHeader(data.size());
```
rjenkinsjr: You are packing the length of the array, but not the array itself. You need to iterate through each item in the array and pack it, presumably by recursively constructing `Datapoint` objects, getting their underlying `Value` and making an array of `Value` objects.

src/main/java/io/auklet/Datapoint.java
```diff
@@ -0,0 +1,246 @@
...
+     *
+     * @param data data to be packed
+     * @throws AukletException to wrap any underlying exceptions.
+     */
+    public Datapoint(Map data) throws AukletException {
+        synchronized (this.msgpack) {
+            this.msgpack.clear();
+            try {
+                this.msgpack.packMapHeader(1)
+                        .packString(this.dataType).packMapHeader(data.size());
```
rjenkinsjr: 1. `Map` is the only Java data type that should not use `this.dataType` (aka. `SIMPLE_KEY`) because we want to logically copy the entire map as-is.
1. You are packing the length of the map, but not the map itself. You need to iterate through each item in the map and pack the keys and values:
   - Keys must be packed as strings. Even though MessagePack supports non-string keys, our data representation in ElasticSearch does not, so we must make them all strings.
   - Values would be packed presumably by recursively constructing `Datapoint` objects, getting their underlying `Value` and making an array of `Value` objects.

src/main/java/io/auklet/sink/AbstractSink.java
```diff
@@ -72,6 +72,27 @@
...
+            this.msgpack.clear();
+            try {
+                this.initMessage(11);
+                this.msgpack
+                        .packString("timestamp").packLong(System.currentTimeMillis())
+                        // User defined type
+                        .packString("type").packString(dataType)
+                        .packString("payload").packArrayHeader(1)
+                        .packBinaryHeader(datapoint.dataValue.length)
+                        .addPayload(datapoint.dataValue);
```
rjenkinsjr: Per my overall comment, I believe this approach will marshal the data incorrectly. The new `Datapoint.getValue().writeTo(this.msgpack)` method should be used here.

src/main/java/io/auklet/Auklet.java
```diff
@@ -262,6 +265,38 @@ public static void send(@Nullable final Throwable throwable) {
         }
     }
 
+    /**
+     * <p>Sends the given string of dataString and dataType as a <i>datapoint</i>.</p>
+     *
+     * @param dataType User specified string which describes the provided data
+     * @param datapoint Class containing serialized user data for mqtt transmission
```
rjenkinsjr: This description needs to be updated. Something like "any value. `null`, primitives/boxed primitives, `BigInteger`s, `BigDecimal`s, `List`s and `Map`s are all serialized into appropriate JSON-like structures. All other object types are converted to a string via `toString()`."

src/main/java/io/auklet/Auklet.java
```diff
@@ -504,6 +539,28 @@ private void doSend(@Nullable final Throwable throwable) {
...
+     * <p>Queues a task to submit the given datapoint to the data sink.</p>
+     * @param data
+     * @param dataType
+     */
+    private void doSend(@NonNull final String dataType, @NonNull final Datapoint data) {
+        try {
+            this.scheduleOneShotTask(new Runnable() {
+                @Override public void run() {
+                    try {
+                        LOGGER.debug("Sending datapoint: {}", dataType);
```
rjenkinsjr: Let's add the `data.toString()` call here as part of this PR so we don't have to do another PR just for that.

src/main/java/io/auklet/core/Datapoint.java
```diff
@@ -0,0 +1,204 @@
...
+import org.msgpack.value.Value;
+import org.msgpack.value.ImmutableValue;
+import org.msgpack.value.ValueFactory;
+import net.jcip.annotations.Immutable;
+
+import java.io.*;
+import java.util.*;
+
+/**
+ *
```
rjenkinsjr: Needs Javadoc. Presumably we have some language-agnostic description of Auklet datapoints that we can use here.

src/main/java/io/auklet/core/Datapoint.java
```diff
@@ -0,0 +1,204 @@
...
+        return this.value;
+    }
+
+    /**
+     * <p>Currently not utilized, will evnetually return a string representation of the datapoint for logging purposes</p>
+     *
+     * @return Empty string
+     */
+    public String toString() {
+        return "";
```
rjenkinsjr: Let's implement this as follows:
- Add a new `private final String asString` class member and return it from this method. `String`s are immutable so no need to copy it.
- In the `convertData` method (see my other comment) construct a string representation for each if/switch possibility. For example, `Integer` and `int` values should be converted via `Integer.toString(data)`.

src/main/java/io/auklet/core/Datapoint.java
```diff
@@ -0,0 +1,204 @@
...
+    private static String SIMPLE_KEY = "value";
+    private final ImmutableValue value;
+
+    /**
+     * <p>Initializes a Datapoint class which packs up the passed data into messagepack byte arrays</p>
+     *
+     * @param data data to be packed
+     * @throws AukletException to wrap any underlying exceptions.
+     */
+    public Datapoint(@Nullable Object data) throws AukletException {
```
rjenkinsjr: The "silhouette" of this constructor is an opportunity for collapsing of some duplicate code.

1. Make a private method `@NonNull private static Map<String, Value> convertData(@Nullable Object data)` that returns the appropriate `ValueFactory.newSomething` based on all the if/switch logic we already have, (or throws an exception if appropriate). Always return a `Map` object with one entry, whose key is the string representation (see my comment on `toString`) and whose value is the `Value` object.
1. Rewrite the constructor so that it:
    - First checks if `data instanceof Datapoint` and if so reuses its value.
    - Otherwise, passes `data` to `convertData` and uses its non-null value as shown below.

```
public Datapoint(Object data) {
  if (data instanceof Datapoint) {
    Datapoint newDatapoint = (Datapoint) data;
    this.value = newDatapoint.getValue();
    this.asString = newDatapoint.toString();
  } else {
    Map.Entry<String, Value> converted = Datapoint.convertData(data).entrySet().iterator().next();
    this.value = ValueFactory.newMap(
        ValueFactory.newString(Datapoint.SIMPLE_KEY),
        converted.getValue()
    );
    this.asString = converted.getKey();
  }
}
```

src/main/java/io/auklet/sink/AbstractSink.java
```diff
@@ -52,7 +53,7 @@
                         .packString("timestamp").packLong(System.currentTimeMillis())
                         .packString("excType").packString(throwable.getClass().getName())
                         .packString("message").packString(Util.orElse(throwable.getMessage(), ""))
-                        .packString("stackTrace").packArrayHeader(stackTrace.length);
+                         .packString("stackTrace").packArrayHeader(stackTrace.length);
```
rjenkinsjr: Extraneous whitespace.

src/main/java/io/auklet/sink/AbstractSink.java
```diff
@@ -68,17 +69,39 @@
             }
             byte[] payload = this.msgpack.toByteArray();
             if (payload == null || payload.length == 0) return;
-            this.write(payload);
+            this.write(payload, this.getAgent().getDeviceAuth().getMqttEventsTopic());
+        }
+    }
+
+    public void send(@NonNull String dataType, @NonNull Datapoint datapoint) throws AukletException {
```
rjenkinsjr: For consistency, this method needs to `@Override` a corresponding declaration in the `Sink` interface. That interface declaration will need a Javadoc; the existing `send` method Javadoc should suffice for inspiration.

This will also ensure that an implementation for `SerialPortSink` is developed, since it is currently missing.

src/main/java/io/auklet/sink/AbstractSink.java
```diff
@@ -68,17 +69,39 @@
...
+            if (payload == null || payload.length == 0) return;
+            this.write(payload, this.getAgent().getDeviceAuth().getMqttDatapointsTopic());
         }
     }
 
     /**
      * <p>Writes the given byte array to the underlying data sink.</p>
      *
      * @param bytes the byte array, never {@code null} or empty.
+     * @param topic the topic to send to, never {@code null} or empty.
```
rjenkinsjr: Need to state here that not all implementations need to use this value, since `SerialPortSink` puts the topic into the payload during its `send` method.

src/main/java/io/auklet/core/Datapoint.java
```diff
@@ -199,6 +158,6 @@ public Datapoint(@Nullable Object data) throws AukletException {
      * @return Empty string
      */
     public String toString() {
```
rjenkinsjr: Need to update these Javadocs and annotate the method as `@NonNull`

src/main/java/io/auklet/core/Datapoint.java
```diff
@@ -16,13 +16,14 @@
...
 
 /**
- *
+ * <p>Datapoints are a way to log your own custom data to the auklet backend.</p>
  */
 @Immutable
 public final class Datapoint {
 
-    private static String SIMPLE_KEY = "value";
+    private static String simpleKey = "value";
```
rjenkinsjr: This should be `private static final` and should use the previous all-capitals name.

src/main/java/io/auklet/sink/AbstractSink.java
```diff
@@ -98,7 +98,7 @@ public void send(@NonNull String dataType, @NonNull Datapoint datapoint) throws
      * <p>Writes the given byte array to the underlying data sink.</p>
      *
      * @param bytes the byte array, never {@code null} or empty.
-     * @param topic the topic to send to, never {@code null} or empty.
+     * @param topic the topic to send to, never {@code null} or empty. Not all methods require this parameter.
```
rjenkinsjr: For clarity let's say `Implementations are not required to use this parameter.`

src/main/java/io/auklet/core/Datapoint.java
tylerwiersing: CC complains about the all capital name, we could ignore it though

rjenkinsjr: That's odd, it shouldn't. `private static final`s are considered "constants" in Java and conventionally use all uppercase. Let's make the change anyway and see what it says.

src/main/java/io/auklet/core/Datapoint.java
```diff
@@ -21,28 +19,33 @@
 @Immutable
 public final class Datapoint {
 
-    private static String simpleKey = "value";
+    private static final String SIMPLE_KEY = "value";
     private final ImmutableValue value;
     private final String asString;
 
+    public final String dataType;
```
rjenkinsjr: Needs to be `private`.

src/main/java/io/auklet/core/Datapoint.java
```diff
@@ -108,23 +111,26 @@ public Datapoint(@Nullable Object data) throws AukletException {
                     ValueFactory.newBoolean((boolean) data));
         } else if (data.getClass().isArray() || data instanceof List) {
             // construct new array of values from looping through all members
-            List<ImmutableValue> newDatapoints = new ArrayList<>();
+            List<Value> newDatapoints = new ArrayList<>();
```
rjenkinsjr: Why did this change from `ImmutableValue`? Do we need to make the `convertData` method return `Map<String, ImmutableValue>` so we can switch this back?

tylerwiersing: Yep we'd need to make the method return an `ImmutableValue` since it cant convert `Value` to `ImmutableValue`

src/main/java/io/auklet/Auklet.java
```diff
@@ -262,6 +266,42 @@ public static void send(@Nullable final Throwable throwable) {
...
+     * <p>Sends the given string of dataString and dataType as a <i>datapoint</i>.</p>
+     *
+     * @param dataType User specified string which describes the provided data
+     * @param datapoint Any value. null, primitives/boxed primitives,
+     *                  BigIntegers, BigDecimals, Lists and Maps are all
+     *                  serialized into appropriate JSON-like structures.
+     *                  All other object types are rejected and cause an
+     *                  AukletException.
+     */
+    public static void send(final String dataType, final Object datapoint) {
```
rjenkinsjr: `dataType` should be `@NonNull` and `datapoint` should be `@Nullable`

src/main/java/io/auklet/core/Datapoint.java
```diff
@@ -0,0 +1,175 @@
...
+    private final String dataType;
+
+    /**
+     * <p>Initializes a Datapoint class which packs up the passed data into messagepack byte arrays</p>
+     *
+     * @param dataType string defining the type of the data to be packed
+     * @param data data to be packed
+     * @throws AukletException to wrap any underlying exceptions.
+     */
+    public Datapoint(@NonNull String dataType, @Nullable Object data) throws AukletException {
```
rjenkinsjr: `dataType` needs a null check before everything else and should throw an `AukletException` if it is null.

src/main/java/io/auklet/core/Datapoint.java
```diff
@@ -0,0 +1,175 @@
...
+    @NonNull public ImmutableValue getValue() {
+        return this.value;
+    }
+
+    /**
+     * <p>Returns the string representation of the value stored in the datapoint</p>
+     *
+     * @return String representation of the current value of the datapoint
+     */
+    @NonNull public String toString() {
```
rjenkinsjr: This needs an `@Override` annotation, in addition to `@NonNull`.

src/main/java/io/auklet/sink/AbstractSink.java
```diff
@@ -68,17 +69,40 @@
             }
             byte[] payload = this.msgpack.toByteArray();
             if (payload == null || payload.length == 0) return;
-            this.write(payload);
+            this.write(payload, this.getAgent().getDeviceAuth().getMqttEventsTopic());
+        }
+    }
+
+    @Override public void send(@NonNull Datapoint datapoint) throws AukletException {
```
rjenkinsjr: Need to check `datapoint == null` before everything else and throw an `AukletException` if true.

PR #74, opened 2019-04-25
CPU and Memory Info Update

tylerwiersing:
Update to allow for the gathering of CPU and Memory information without the presence of certain java apis.

### DIFF
--- a/src/main/java/io/auklet/platform/metrics/AbstractMetrics.java
+++ b/src/main/java/io/auklet/platform/metrics/AbstractMetrics.java
@@ -1,0 +1,82 @@
+package io.auklet.platform.metrics;
+
+import android.app.ActivityManager;
+import android.content.Context;
+import android.os.Build;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import io.auklet.AukletException;
+import net.jcip.annotations.GuardedBy;
+import net.jcip.annotations.ThreadSafe;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+
+
+public abstract class AbstractMetrics {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractMetrics.class);
+
+    public final Object lock = new Object();
+    @GuardedBy("lock") private long total = 0L;
+    @GuardedBy("lock") private long totalBefore = 0L;
+    @GuardedBy("lock") private long totalDiff = 0L;
+    @GuardedBy("lock") private long work = 0L;
+    @GuardedBy("lock") private long workBefore = 0L;
+    @GuardedBy("lock") private long workDiff = 0L;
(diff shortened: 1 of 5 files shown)


src/main/java/io/auklet/platform/metrics/JavaMetrics.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * <p>Returns the CPU usage of the OS on which this agent is running.</p>
+     *
+     * @return a non-negative value. If running on Android 8 or higher, will always be zero.
+     */
+     @Override public float getCpuUsage() {
+        long processors = OSMX.BEAN.getAvailableProcessors();
+        if (processors == 0) {
+            Runnable runnableCPU = this.calculateCpuUsage();
```
rjenkinsjr: This scheduling of the repeating CPU usage task, if required, should happen when this object is constructed, not every time `getCpuUsage` is called.

src/main/java/io/auklet/platform/metrics/JavaMetrics.java
```diff
@@ -0,0 +1,96 @@
...
+import java.io.InputStreamReader;
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+
+
+/** <p>Platform methods specific to Java SE (and variants).</p> */
+public final class JavaMetrics extends AbstractMetrics {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(JavaMetrics.class);
+    public static Auklet agent;
```
rjenkinsjr: This field does not appear to be used.

src/main/java/io/auklet/platform/metrics/AbstractMetrics.java
```diff
@@ -0,0 +1,77 @@
...
+     * <p>Returns a runnable task that periodically gets system CPU usage from the
+     * Android device on which the agent is running. Android has no APIs available to
+     * obtain this information, so we have to use a background thread to read the
+     * {@code /proc/stat} file.</p>
+     *
+     * @return {@code null} iff running on Android 8 or higher, in which case no
+     * background task will be executed.
+     */
+    @Nullable public Runnable calculateCpuUsage() {
+        if (Build.VERSION.SDK_INT >= 26) return null;
```
rjenkinsjr: This logic is Android-specific and should go into the `AndroidMetrics` constructor. Thus, this method should never return `null` and should only be invoked when conditions demand it (e.g. this conditional for Android, when `OSMX.getAvailableProcessors()` returns 0 on Java).

src/main/java/io/auklet/platform/metrics/AbstractMetrics.java
```diff
@@ -0,0 +1,77 @@
...
+import net.jcip.annotations.GuardedBy;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+
+
+public abstract class AbstractMetrics {
```
rjenkinsjr: This class and it descendants should be Javadoc'd and marked `@ThreadSafe`.

PR #78, opened 2019-10-28
Improve Auklet API credential security

rjenkinsjr:
This PR adds better handling of requests to the Auklet API by ensuring that any usage of the API key is tied exclusively to requests that only use the base URL. This ensure that API keys will only ever be sent to A. `api.auklet.io`, or B. whatever value the end user has configured for the base URL, in which case the security of their API key becomes their responsibility.


shogun656:
Can you also look at the one issue that codeclimate has raised

